PR #95, opened 2019-08-07
Definitions of twitter (preview) (step 1)

Misaka-0x447f:


src/utils/utils.ts
```diff
@@ -41,3 +46,5 @@ export function untilDocumentReady() {
         document.addEventListener('readystatechange', resolve, { once: true, passive: true })
     })
 }
+
+export const nop: (...args: any[]) => any = () => {};
```
Jack-Works:
```ts
export { noop } from 'lodash-es'
```

Misaka-0x447f: noop, which has a type ```void```, is not a compatible with type ```() => void```

Misaka-0x447f:
This pull-request is still not functional but should not affect functional of Facebook worker/
Just make changes on this up to date.

src/utils/assert.ts
```diff
@@ -0,0 +1,13 @@
+export class counter {
```
Jack-Works: Maybe a better name?

src/social-network/defaults/injectCommentBox.tsx
```diff
@@ -5,9 +5,23 @@ import { useValueRef } from '../../utils/hooks/useValueRef'
...
 import Services from '../../extension/service'
 import { renderInShadowRoot } from '../../utils/jss/renderInShadowRoot'
+import { dispatchCustomEvents, selectElementContents, sleep } from '../../utils/utils'
 
-export function injectCommentBoxDefault(onPasteToCommentBox: (encryptedComment: string, current: PostInfo) => void) {
+export function injectCommentBoxDefault(
+    onPasteToCommentBox: (
+        encryptedComment: string,
+        current: PostInfo,
+    ) => void = async function onPasteToCommentBoxFacebook(encryptedComment, current) {
```
Jack-Works: Why should `onPasteToCommentBox` have a default value and it called `onPasteToCommentBoxFacebook`? This should be platform-specific

src/social-network-provider/twitter.com/worker/fetch.ts
```diff
@@ -0,0 +1,30 @@
+import { PersonIdentifier, PostIdentifier } from '../../../database/type'
+import { toPostUrl, toProfileUrl } from '../utils/url'
+import tasks from '../../../extension/content-script/tasks'
+import { bioQueryString } from '../utils/selector'
+
+export const fetchPostContent = async (post: PostIdentifier<PersonIdentifier>) => {
+    try {
+        const d = await parseDocument(toPostUrl(post))
+        const content = d.innerText.match(/(🔒.+🔒)/)
```
Jack-Works: You should take care about the key comes from. If some one commented a key under a normal post, it will be a false-positive and it is dangerous.

src/social-network-provider/twitter.com/utils/user.ts
```diff
@@ -0,0 +1,12 @@
+import { SocialNetworkWorkerAndUI } from '../../../social-network/shared'
+import { isNull } from 'lodash-es'
+
+export const usernameValidator: SocialNetworkWorkerAndUI['isValidUsername'] = (name: string) => {
```
Jack-Works: Any reference at twitter for this function?

src/social-network-provider/twitter.com/ui/task.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+    await sleep(400)
+    try {
+        const [i] = editProfileTextareaSelector.evaluateOnce()
+        await sleep(200)
+        i.focus()
+        dispatchCustomEvents('input', i.value + text)
+    } catch {
+        console.warn('Text not pasted to the text area')
+        prompt(geti18nString('automation_request_paste_into_bio_box'), text)
```
Jack-Works: You should detect by something like `innerText`

src/social-network-provider/facebook.com/UI/injectPostInspector.tsx
```diff
@@ -3,7 +3,7 @@ import { isMobileFacebook } from '../isMobile'
 import { PostInfo } from '../../../social-network/ui'
 import { injectPostInspectorDefault } from '../../../social-network/defaults/injectPostInspector'
 
-const defaultBehavior = injectPostInspectorDefault({
+export const defaultBehavior = injectPostInspectorDefault({
```
Jack-Works: Used where?

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,131 @@
...
+import { deconstructPayload } from '../../../utils/type-transform/Payload'
+import { regexMatch, timeout } from '../../../utils/utils'
+import { hasPostPopup } from '../utils/status'
+
+export const resolveLastRecognizedIdentity = (self: SocialNetworkUI) => {
+    const selfSelector = selfInfoSelectors.screenName
+    const assign = () => {
+        const ref = self.lastRecognizedIdentity
+        const i = new PersonIdentifier(host, selfInfoSelectors.screenName.evaluateOnce()[0])
+        if (i.isUnknown) throw new Error('failed to recognize user id')
```
Jack-Works: This function expects to fail to collect info, maybe don't thorw is a good idea

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,131 @@
...
+export const resolveInfoFromBioCard = () => {
+    const userAvatarUrl = bioCard
+        .nth(0)
+        .querySelector<HTMLImageElement>('img')
+        .evaluateOnce()[0].src
+    const userNames = bioCard
+        .nth(1)
+        .evaluateOnce()[0]
+        .innerText.split('\n')
+    const userBio = bioCard.nth(2).evaluateOnce()[0].innerText
```
Jack-Works: You should call `.clone()` before use a shared LiveSelector. LiveSelector has internal state, and any function chains excepts `evaluateOnce` will leads to the changes of the internal state.

And `evaluateOnce` is deprecated, use `evaluate` instead

src/social-network-provider/facebook.com/UI/injectPostInspector.tsx
Misaka-0x447f: image

Jack-Works: If it is only used in facebook support, you should name it as defaultBehaviorFacebook to avoid importing by mistakes. If it is used as a global default, don't do it.

src/social-network/defaults/injectCommentBox.tsx
Misaka-0x447f: fixed in 82e82d5

src/utils/assert.ts
Misaka-0x447f: done

src/social-network-provider/twitter.com/worker/fetch.ts
Misaka-0x447f: This also happens at
https://github.com/DimensionDev/Maskbook/blob/1c2f6c4691eaa015298be8667393fdbbfec1f2da/src/social-network-provider/facebook.com/Worker/fetchPostContent.ts#L22
Maybe refactor this first or... whatever, we need a HTML document parser.

src/social-network-provider/twitter.com/utils/user.ts
Misaka-0x447f: image

src/social-network-provider/twitter.com/ui/task.ts
Misaka-0x447f: TODO added, complete logic later.

src/social-network-provider/twitter.com/ui/fetch.ts
Misaka-0x447f: a06055dc

src/social-network-provider/facebook.com/UI/injectPostInspector.tsx
Misaka-0x447f: already set to nop on the twitter side.

src/social-network-provider/twitter.com/ui/fetch.ts
Misaka-0x447f: TS2339: Property 'evaluate' does not exist on type 'LiveSelector<HTMLImageElement, true>'.

Misaka-0x447f: Ok your package from github is not auto upgrading. ```yarn upgrade``` required. Now a ```just``` script will do this at post install time. We will need npm managing version of this package ASAP.

Misaka-0x447f: 8ec9f6c5

Misaka-0x447f:
Intend: wait until at least one element available
Waiting for verify if it works.
image

src/social-network-provider/twitter.com/ui/fetch.ts
Misaka-0x447f: f1882da7. Now always get new selector. check it out.

just.config.js
```diff
@@ -22,8 +22,10 @@ task('storybook', () => parallel('lint/fix', 'storybook/serve'))
 task('storybook/serve', () => step('start-storybook -p 9009 -s public --quiet', { withWarn: true }))
 task('storybook/build', () => step('build-storybook -s public --quiet', { withWarn: true }))
 
+task('install', () => series('install/holoflows'))
 task('install/holoflows', async () => {
     const base = path.join(process.cwd(), 'node_modules/@holoflows')
+    await step('yarn upgrade @holoflows/kit')
```
Jack-Works: Hmm.. In old days we install @holoflows/kit by `yarn link`. That enable us to develop these two libraries in the same time.

But now @holoflows/kit is more stable than the old days, so month ago I make it as a dependency installed from git.

If kit gets more stable, we may install it directly from npm in the future.

src/social-network/defaults/injectCommentBox.tsx
Jack-Works: Yeah I see but I think `root.querySelector('[contenteditable]')` (some website may use textarea or input) and `dispatchCustomEvents('paste', encryptedComment)` (not every website are using React and listenen from `document`) is not a good default.

src/social-network-provider/twitter.com/worker/fetch.ts
Jack-Works: There is a build-in parser. Checkout http://mdn.io/DOMParser and we're already using it.

src/social-network-provider/twitter.com/utils/user.ts
Jack-Works: Uh... I mean the `name policy` in the twitter.com. Where do you find this rule?

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,155 @@
...
+            const ref = self.lastRecognizedIdentity
+            const info = selfInfoSelectors().screenName.evaluate()
+            const id = new PersonIdentifier(host, notEmpty(info, 'user id not found')[0])
+            equal(id.isUnknown, false, 'user id not recognized')
+            ref.value = { identifier: id }
+        } catch (e) {
+            /**
+             * This function expects to fail to collect info,
+             * maybe don't throw is a good idea
+             *                                  -- Jack-Works
```
Jack-Works: ?????????

src/social-network-provider/facebook.com/ui-provider.ts
```diff
@@ -65,3 +55,17 @@ defineSocialNetworkUI({
     taskGetPostContent: getPostContentFacebook,
     taskGetProfile: getProfileFacebook,
 })
+
+export const updaterFactory = (self: SocialNetworkUI) => {
```
Jack-Works: It also used in twitter support. Please move it out of the facebook support.

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,155 @@
...
+import Services from '../../../extension/service'
+import { PersonIdentifier } from '../../../database/type'
+import { host } from '../index'
+import { getEmptyPostInfo, SocialNetworkUI } from '../../../social-network/ui'
+import { deconstructPayload } from '../../../utils/type-transform/Payload'
+import { regexMatch, timeout } from '../../../utils/utils'
+import { hasPostPopup } from '../utils/status'
+import { equal, notEmpty, notNullable } from '../../../utils/assert'
+
+export const resolveLastRecognizedIdentity = (self: SocialNetworkUI) => {
```
Jack-Works: We should call it resolveLastRecognizedIdentityTwitter to avoid accidentally import at someone else.

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+    new MutationObserverWatcher(selfSelector)
+        .enableSingleMode()
+        .addListener('onAdd', () => assign())
+        .addListener('onChange', () => assign())
+        .startWatch()
+        .then()
+}
+
+export const resolveInfoFromBioCard = () => {
```
Jack-Works: We should call it resolveInfoFromBioCardTwitter to avoid accidentally import at someone else.

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,155 @@
...
+                }).then()
+            }
+            refreshUserInfo()
+            return {
+                onNodeMutation: refreshUserInfo,
+                onTargetChanged: refreshUserInfo,
+            }
+        })
+        .startWatch()
+        .then()
```
Jack-Works: `then()` on the Watcher have side effects. It implments `PromiseLike` but it is not a Promise.

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,155 @@
...
+            collectPostInfo()
+            return {
+                onNodeMutation: collectPostInfo,
+                onTargetChanged: collectPostInfo,
+                onRemove: () => that.posts.delete(proxy),
+            }
+        })
+        .setDomProxyOption({ afterShadowRootInit: { mode: 'closed' } })
+        .startWatch()
+        .then()
```
Jack-Works: same as above

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/**
+ * This can be help to make sure if bioCard exists on the page.
+ * @throws exception if not exist
+ * @return bioCard element, if exists
+ */
+export const fetchBioCard = () => timeout(new MutationObserverWatcher(bioCard()), 10000)
```
Jack-Works: fetchBioCardTwitter

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Test if able to fetch posts.
+ */
+export const fetchPost = async () => {
```
Jack-Works: fetchPostTwitter

src/social-network-provider/twitter.com/ui/fetch.ts
```diff
@@ -0,0 +1,155 @@
...
+    const s = (() => {
+        if (hasPostPopup()) {
+            return postPopupSelector().concat(postsContentSelectors().enableSingleMode())
+        }
+        return postsContentSelectors().enableSingleMode()
+    })()
+    return timeout(new MutationObserverWatcher(s), 10000)
+}
+
+export { registerBioCollector as collectPeople, registerPostCollector as collectPosts }
```
Jack-Works: collectPeopleTwitter
collectPostsTwitter

src/social-network-provider/twitter.com/ui/task.ts
```diff
@@ -0,0 +1,47 @@
+import { dispatchCustomEvents, sleep, untilDocumentReady } from '../../../utils/utils'
+import { editProfileButtonSelector, editProfileTextareaSelector, newPostEditorInnerSelector } from '../utils/selector'
+import { geti18nString } from '../../../utils/i18n'
+import { fetchBioCard, fetchPost, resolveInfoFromBioCard } from './fetch'
+import { notNullable } from '../../../utils/assert'
+
+export const taskPasteIntoPostBox = async (text: string, warningText: string) => {
```
Jack-Works: taskPasteIntoPostBoxTwitter, and also other exports in this file

src/social-network-provider/twitter.com/utils/selector.ts
```diff
@@ -0,0 +1,54 @@
...
+export const postsRootSelector = () => querySelector<HTMLElement>(`[data-testid="primaryColumn"] section`)
+
+export const postPopupSelector = () => querySelector('[aria-labelledby="modal-header"]')
+export const postsSelectors = () => querySelectorAll('article')
+export const postsContentSelectors = () => postsSelectors().querySelectorAll<HTMLElement>(`[lang]`)
+export const fromPostSelectorsSelectPostContentString = '[data-testid="tweet"] > div:nth-of-type(2)'
+
+// self infos
+const base = querySelector<HTMLScriptElement>('#react-root + script')
+const name = /"session":{.*?"user":{.*?"screen_name":"(.*?)","name":"(.*?)"}}/
```
Jack-Works: Uh.. Match by RegExp? It seems like a JSON or something...

src/social-network/defaults/injectCommentBox.tsx
```diff
@@ -5,9 +5,19 @@ import { useValueRef } from '../../utils/hooks/useValueRef'
...
 import Services from '../../extension/service'
 import { renderInShadowRoot } from '../../utils/jss/renderInShadowRoot'
+import { dispatchCustomEvents, selectElementContents, sleep } from '../../utils/utils'
 
-export function injectCommentBoxDefault(onPasteToCommentBox: (encryptedComment: string, current: PostInfo) => void) {
-    return function injectPostBox(current: PostInfo) {
+const defHandler = async (encryptedComment: string, current: PostInfo) => {
+    const root = current.rootNode
+    selectElementContents(root.querySelector('[contenteditable]')!)
+    dispatchCustomEvents('paste', encryptedComment)
```
Jack-Works: Not every website are using React. This shouldn't be the default.

src/social-network/ui.ts
```diff
@@ -57,6 +62,12 @@ interface SocialNetworkUIInformationCollector {
      * ```
      *
      * If `selectedIdentity` is null, you should try to set it to an identity in the `myIdentitiesRef`
+     *
+     */
+    /**
+     * This function expects to fail to collect info,
+     * maybe don't throw is a good idea
+     *                                  -- Jack-Works
```
Jack-Works: ??????????

src/social-network-provider/twitter.com/worker/fetch.ts
Misaka-0x447f: Okay a TODO will be added.

src/social-network/defaults/injectCommentBox.tsx
Misaka-0x447f: Ok but although ```git blame``` is pointing this edit of line to me, this is actually added more earlier and written by you. ......A todo has been added.

just.config.js
Misaka-0x447f: yeah a npm package can be helpful

src/social-network-provider/twitter.com/ui/fetch.ts
Misaka-0x447f: Forgot add a TODO comment here

src/social-network-provider/twitter.com/utils/selector.ts
Misaka-0x447f: Match by RegExp could be faster than parsing the whole JSON in case I only want one value in it.

src/social-network/ui.ts
Misaka-0x447f: Removed signature.

src/social-network/defaults/injectCommentBox.tsx
Misaka-0x447f: need a todo...?

src/social-network-provider/twitter.com/utils/user.ts
Misaka-0x447f: Sure, check it out. https://help.twitter.com/en/managing-your-account/twitter-username-rules

PR #99, opened 2019-08-09
Chore related to option page's UI

SunriseFox:
This PR includes:
+ extracted nav component
+ on clicking close/finish button, close the popup dialog instead of closing the tab
+ close the vide o track when it should be

Please test:
+ mobile

### DIFF
--- a/src/components/Welcomes/Navigation/Navigation.tsx
+++ b/src/components/Welcomes/Navigation/Navigation.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react'
+
+import Close from '@material-ui/icons/Close'
+import ArrowBack from '@material-ui/icons/ArrowBack'
+import { geti18nString } from '../../../utils/i18n'
+import { makeStyles, Button } from '@material-ui/core'
+
+interface Props {
+    back?: () => void
+    close?: () => void
+}
+const useStyles = makeStyles(theme => ({
+    nav: {
+        paddingTop: theme.spacing(1),
+        paddingLeft: theme.spacing(1),
+        display: 'flex',
+    },
+    navBackButton: {
+        color: theme.palette.text.hint,
+        marginLeft: theme.spacing(1),
+        marginRight: 'auto',
+    },
+    navCloseButton: {
+        color: theme.palette.text.hint,
+        marginRight: theme.spacing(1),
+        marginLeft: 'auto',
+    },
+    navBackIcon: {
+        marginRight: theme.spacing(1),
+    },
+    navCloseIcon: {
+        marginLeft: theme.spacing(1),
+    },
+}))
+export default function Navigation({ back, close }: Props) {
+    const classes = useStyles()
+    return (
+        <nav className={classes.nav}>
+            {back ? (
+                <Button onClick={back} disableFocusRipple disableRipple className={classes.navBackButton}>
+                    <ArrowBack className={classes.navBackIcon} />
+                    {geti18nString('back')}
+                </Button>
+            ) : (
+                <></>
+            )}
+            {close ? (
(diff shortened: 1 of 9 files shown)


src/extension/options-page/Welcome/index.tsx
```diff
@@ -68,7 +68,7 @@ const WelcomeActions = {
         this.autoVerifyBio(user, prove)
     },
     onFinish(reason: 'quit' | 'done') {
-        window.close()
+        ;((this as unknown) as { props: RouteComponentProps }).props.history.replace('/')
```
Jack-Works: Why?

src/extension/options-page/Welcome/index.tsx
```diff
@@ -148,6 +148,13 @@ export default withRouter(function _WelcomePortal(props: RouteComponentProps<{ i
     const provePost = useValueRef(ProvePostRef)
     const identifier = useValueRef(IdentifierRef)
 
+    const onFinish = WelcomeActions.onFinish.bind({ props })
+    const closeDialog = (e: React.MouseEvent) => {
```
Jack-Works: I don't think it's the right way to close a dialog. This dialog is by-design to be always open, if you want to close it, just change the router to `/`.

src/extension/options-page/Welcome/index.tsx
```diff
@@ -158,14 +165,14 @@ export default withRouter(function _WelcomePortal(props: RouteComponentProps<{ i
     }, [props.location.search])
 
     return (
-        <ResponsiveDialog open>
+        <ResponsiveDialog open onClick={closeDialog}>
```
Jack-Works: do you mean `onClose`?

src/extension/options-page/Welcome/index.tsx
Jack-Works: And we have a close button, why detect target source?

src/extension/options-page/Welcome/index.tsx
Jack-Works: You can change the parameter of this function.

src/components/Welcomes/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const classes = useStyles()
+    return (
+        <nav className={classes.nav}>
+            {back ? (
+                <Button onClick={back} disableFocusRipple disableRipple className={classes.navBackButton}>
+                    <ArrowBack className={classes.navBackIcon} />
+                    {geti18nString('back')}
+                </Button>
+            ) : (
+                <></>
```
Jack-Works: You can use `null` here >_>

src/extension/options-page/Welcome/index.tsx
SunriseFox: Using `this` is quite common these days... maybe. So how to make typescript recognize `this` the graceful way?

src/extension/options-page/Welcome/index.tsx
```diff
@@ -148,6 +153,8 @@ export default withRouter(function _WelcomePortal(props: RouteComponentProps<{ i
     const provePost = useValueRef(ProvePostRef)
     const identifier = useValueRef(IdentifierRef)
 
+    const onFinish = WelcomeActions.onFinish.bind({ props })
```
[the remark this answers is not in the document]
SunriseFox: I tried to replace all `window.close` with this method as the tab window cannot be closed when using firefox (and currently there's no way to detect whether the window.close option succeeds).

Maybe we could add another function argument to pass the `props` or something? In fact I don't quite know how to do it the React way... I was just trying to take the React context into the function like Vue does.

You may modify it as your wish.

Jack-Works:
Rebased master

PR #108, opened 2019-08-15
Fix/chore

SunriseFox:
This PR solves the following issues:

+ html-webpack-plugin inject unnecessary chunks into index.html (which prevents popup from being loaded)
+ throws on `[undefined]` when no identity is found.

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -32,6 +32,13 @@ module.exports = function override(/** @type{import("webpack").Configuration} */
     config.module.exprContextCritical = false
     config.module.unknownContextCritical = false
 
+    // Prevent all other chunks from being injected to index.html
+    config.plugins.forEach(p => {
+        if (p.constructor.name !== 'HtmlWebpackPlugin') return
+        const {devtools, app, ...exclude} = config.entry
+        Object.keys(exclude).forEach(e => p.options.excludeChunks.push(e))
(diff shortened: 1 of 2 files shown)


src/components/Welcomes/1a1a.tsx
```diff
@@ -28,7 +28,7 @@ const useStyles = makeStyles(theme => ({
 }))
 export default function Welcome({ next, identities, didntFindAccount }: Props) {
     const classes = useStyles()
-    const [selected, setSelect] = useState<Person[]>([identities[0]])
+    const [selected, setSelect] = useState<Person[]>(identities[0] ? [identities[0]] : [])
```
Jack-Works:
```ts
const [selected, setSelect] = useState<Person[]>(identities[0] ? identities : [])
```

config-overrides.js
```diff
@@ -32,6 +32,12 @@ module.exports = function override(/** @type{import("webpack").Configuration} */
     config.module.exprContextCritical = false
     config.module.unknownContextCritical = false
 
+    config.plugins.forEach(p => {
```
Jack-Works: Can you leave a comment to describe what does this code do?

SunriseFox: I think it quite obvious...

PR #119, opened 2019-08-21
development script update

Misaka-0x447f:
Limited by extension, HMR not available for now. You will need a F5 to reload the extension,


Misaka-0x447f:
TRY this on your env before merge

Misaka-0x447f:
This project has a integrated auto content script update and don't need another one. And background script update may be annoying since some editor user will got unexpected auto update.

PR #121, opened 2019-08-22
fix bugs and add new features

### LINKED ISSUE #79: [High Priority]Several on boarding problems
Priority: **Extremely High** for various purposes.

After new user download the json key file and trying to post the pub key:

- If the auto posting function failed, now we require the user to copy & paste manually; We don't help user copy the pub key, and if the user **close the interface before copying the pub key**, they have to redo the whole on boarding process again (creating new key pairs), which is very user unfriendly.

Potential Fixes:

- [x]  Automatically **Copy** the pub key once it's time for posting pub key etc -> @Jack-Works 

- For power user, they may try to find the pub key in the json file since they know what's key pairs. They will not able to find (or not easily find) the pub key/proof post pub key in the json file. They will think we are brain dead and doubt the system design. 
> So let's change the design of key pairs/json files, since it's better to put that pub key info in the key file?
- Still, user (even power user) will forget to post their bio/proof post 1st and post their encrypted post before that. 
>  Screen Shot 2019-08-01 at 18 28 08

Do we have any way on the UX/UI side to prevent this from happening? @neruthes

### LINKED ISSUE #118: ux: new url permission scheme does not work with profile import
importing an old profile does not trigger the webpage access permission request, which is needed since d9f1f31. someone should write a "smarter" handler for imported personae.

Jack-Works:
Fix 10 bugs. (see commit log)

Feature:
- request permission when backup. close #118
- support decrypt comments on Facebook mobile
- send encrypt comment from mobile facebook
- add a hint in json for power users, related #79
- try to copy to clipboard when user clicks verify. close #79
- support dark mode in options page


src/utils/type-transform/BackupFile.ts
```diff
@@ -1,7 +1,24 @@
...
+// Before 8/21/2019, we only support facebook, so we can auto upgrade the backup file
+const facebookHost = ['https://m.facebook.com/*', 'https://www.facebook.com/*']
+/**
+ * Refer to https://github.com/DimensionDev/Maskbook/issues/79
+ * Some power users may try to find prove post in the JSON file.
+ *
+ * We should give them a hint.
+ */
+export const JSON_HINT_FOR_POWER_USER =
+    'Power Users look here! You should add the value in this field to the bio to make Maskbook works! ->'
```
Artoria2e5: work*

src/content-script.ts
```diff
@@ -1,11 +1,11 @@
 const public_path = (Reflect.get(window, 'browser') || Reflect.get(window, 'chrome')).runtime.getURL('/')
 __webpack_public_path__ = public_path
 const orig = Object.getOwnPropertyDescriptor(HTMLScriptElement, 'src')!
-const chunks = new WeakSet<Node>()
+const __webpack__loading__chunks__ = new Set<Node>()
```
Artoria2e5: we are... taking responsibility know? should we also delete it in a `catch` later or?

src/extension/background-script/WelcomeService.ts
```diff
@@ -44,6 +45,9 @@ async function generateBackupJSON(whoAmI: PersonIdentifier, full = false): Promi
             localKey: await exportKey(localKeys.get(data.identifier.network)![data.identifier.userId]!),
             publicKey: await exportKey(data.publicKey),
             privateKey: await exportKey(data.privateKey),
+            [JSON_HINT_FOR_POWER_USER]:
+                (await getMyProveBio(data.identifier)) ||
+                'We are sorry, but this field is not available, you setup Maskbook again may help.',
```
Artoria2e5: `available. It may help to set up Maskbook again.`

src/content-script.ts
```diff
@@ -17,28 +17,23 @@ Object.defineProperty(HTMLScriptElement.prototype, 'src', {
...
         if (val.startsWith(public_path)) {
-            console.log('Hijacking', val)
+            console.debug('Hijacking', val)
+            const err = () => (this.onerror || (() => {})(), '')
             fetch(val)
-                .then(data => data.text())
-                .then(eval)
-            chunks.add(this)
+                .then(data => data.text(), err)
+                .then(eval, err)
```
Artoria2e5: FYI, the current es/tslintrc hates `eval`. May be a good idea to add a comment to shut it up.

src/content-script.ts
Jack-Works: yeah, webpack's default way doesn't run our code in the content script context. There is a way to not use eval. Let background script do this but that's unnecessary complicated.

I prefer run es module directly but iOS won't work with it. Upstream (realms-shim) block us.

Artoria2e5: We can do it in webpack! See #125.

Jack-Works:
@Artoria2e5 @SunriseFox @Misaka-0x447f @guanbinrui any reviewers?

SunriseFox:
We need to talk... We need to properly merge the current changes before moving to next. The two PRs' base codes (webpack configuration, linter configuration, etc.) differ with their base master branch and the current master branch, and they are making changes on the same section of codes. So we'd better sort them out.

Jack-Works:
I will rebase to the master if any other prs get merged so there will be no conflicts.

Jack-Works:
@SunriseFox all problems are fixed.

any new review?

PR #123, opened 2019-08-22
chore(lint): use husky to lint code and commit message

SunriseFox:


Jack-Works:
approved

Do you mind to add check to ci file?

What about add and enforce tslint/eslint?

SunriseFox:
> approved
> 
> Do you mind to add check to ci file?
> 
> What about add and enforce tslint/eslint?

Maybe we need to bring eslint-loader back. This shall be done at this weekend, together with other chores with respect to the toolchains.

Jack-Works:
uhh, do we really need a git hook? Since all the codes must be merged by pull request, we can just reject the bad code in CI

What do you think?

SunriseFox:
that seems ok.

Jack-Works:
We will do that in #126

PR #125, opened 2019-08-22
chore: switch to more proper webpack chunking

Artoria2e5:
Why fuck with eval when you can let webpack do it?

PS: the vendor split is deterministic for the entry-point list and the dependencies used by the entry points. WebpackHtmlPlugin could allegedly do that automatically, but guess what IT DOES NOT. I wonder how `index.html` gets it.

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -9,9 +9,10 @@ const polyfills = [
     'node_modules/webextension-polyfill/dist/browser-polyfill.min.js',
     'node_modules/webextension-polyfill/dist/browser-polyfill.min.js.map',
 ]
-const public = path.join(__dirname, './public')
-const publicPolyfill = path.join(__dirname, './public/polyfill')
-const dist = path.join(__dirname, './dist')
+const src = file => path.join(__dirname, file)
+const public = src('./public')
+const publicPolyfill = src('./public/polyfill')
+const dist = src('./dist')
 
 process.env.BROWSER = 'none'
 /**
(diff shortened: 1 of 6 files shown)


Artoria2e5:
oops, the curse of the force-push...

Jack-Works:
Cool. But I prefer the `WebpackHtmlPlugin` way.

Creating a new background page and let webpack inject chunks.

And further, I think we can create a "content script page" to get the dependencies and parse that page in the background page in runtime, we can get rid of thoes hacks on HTMLScriptElement.prototype.src

Artoria2e5:
I already got rid of them

Jack-Works:
Uhh, I found this not working on my machine.

WebCryptoLiner and WebExtension polyfill loaded, no logs printed, I think background service does not started

Artoria2e5:
screenshot the dom tree will ya

Jack-Works:
Nothing new.

```html
<html><head>
<script src="/polyfill/webcrypto-liner.shim.js"></script>
<script src="/polyfill/browser-polyfill.min.js"></script>
<script src="/js/vendors~app~backgroundservice~contentscript.chunk.js"></script><!-- XXX: bit hacky, relies on entry point list, need to read up on webpack plugins -->
<script src="/js/backgroundservice.js"></script>
</head><body>
</body></html>
```

Artoria2e5:
i have no problem with the background, but the content script understandably fails to inject correctly the dependencies. gonna write a plugin for browser.geturl wrapping.

config-overrides.js
```diff
@@ -23,25 +24,41 @@ module.exports = function override(/** @type{import("webpack").Configuration} */
...
+        }),
+        new HtmlWebpackPlugin({
+            chunks: ['options-page'],
+            filename: 'index.html',
+            ...includeWebExtPolyfill,
+        }),
+        new HtmlWebpackPlugin({
+            chunks: ['content-script'],
+            filename: 'generated__content__script.html',
+            ...includeWebExtPolyfill,
```
Misaka-0x447f: Consider make a factory for these duplicated code?

config-overrides.js
```diff
@@ -10,9 +11,9 @@ module.exports = function override(/** @type{import("webpack").Configuration} */
...
     config.entry = {
         devtools: 'react-devtools',
-        app: path.join(__dirname, './src/index.tsx'),
-        contentscript: path.join(__dirname, './src/content-script.ts'),
-        backgroundservice: path.join(__dirname, './src/background-service.ts'),
+        'options-page': path.join(__dirname, './src/index.tsx'),
+        'content-script': path.join(__dirname, './src/content-script.ts'),
+        'background-service': path.join(__dirname, './src/background-service.ts'),
         injectedscript: path.join(__dirname, './src/extension/injected-script/index.ts'),
         qrcode: path.join(__dirname, './src/web-workers/QRCode.ts'),
```
Misaka-0x447f: Consider make a factory for this:
```javascript
(file) => path.join(__dirname, file)
```

PR #126, opened 2019-08-23
Chore/linter renewal

Artoria2e5:
switching to eslint for what tslint used to do


.circleci/config.yml
```diff
@@ -6,7 +6,8 @@ version: 2.1
 jobs:
     build:
         docker:
-            - image: circleci/node:11.0
+            - image: circleci/node:current
+            - image: circleci/node:lts
```
Jack-Works: Why do we need 2 images?

Artoria2e5:
was trying to be "careful" by doing a lts build. turns out it aint good; gonna amend to current lonly

Artoria2e5:
and for fucks sake i really want to read the stderr..

Artoria2e5:
lmao https://circleci.com/gh/DimensionDev/Maskbook/598

now switching to current-only

Artoria2e5:
you know what? our ci node_modules cache is fucked. react script (maybe outdated) wants eslint 5 but our thing has eslint 6.

Artoria2e5:
oh fuck the type signatures are not matching up

Jack-Works:
@Artoria2e5 we should add prettier check and eslint check in CI build step!

Artoria2e5:
The problem with styled `Main` in Dashboard.tsx is probably tracable to mui-org/material-ui@5b25ed3 or mui-org/material-ui@743d7d5. In both cases there appear to be a perfectly reasonable match that handles the themed case but typescript refuses to select it.

Typing changes in material UI is also forcing me to add `<Theme>` to many makeStyles calls. How do I correctly yell at it then?

SunriseFox:
> The problem with styled `Main` in Dashboard.tsx is tracable to [mui-org/material-ui@29b5051](https://github.com/mui-org/material-ui/commit/29b5051381b73162bd72bd7897961727cac604a1). There appears to be a perfectly reasonable match that handles the themed case but typescript refuses to select it.
> 
> Typing changes in material UI is also forcing me to add `<Theme>` to many makeStyles calls. How do I correctly yell at it then?

Maybe we could just remove 
```javascript
const theme = useTheme()
```
and ` theme={theme}`

Artoria2e5:
The fucked up thing is that Dashboard.tsx actually uses something from the theme. A minimal reproduction has been thrown onto mui-org/material-ui#17158.

Jack-Works:
- [ ] Add CI script to let prettier enforced
- [ ] Add CI script to let eslint enforced

Artoria2e5:
Just rebased it onto master. Testing the build now.

Artoria2e5:
great! all errors are deprecation errors!

just.config.js
```diff
@@ -15,8 +15,12 @@ task('react/start', () => step('react-app-rewired start'))
 task('react/build', () => step('react-app-rewired build'))
 task('react/test', () => step('react-app-rewired test'))
 
-task('lint', () => lintCommand('check'))
-task('lint/fix', () => lintCommand('write', 'warn'))
+task('lint', () => prettierCommand('check'))
+task('lint/fix', () => prettierCommand('write', 'warn'))
+task('lint/strictonce', async () => {
+    await step('eslint --ext tsx,ts ./src/ --cache', { withWarn: true })
+    await step('prettier --check "./src/**/*.{ts,tsx}" --loglevel warn ', { withWarn: true })
```
Jack-Works: I think these two tasks should run in parallel so they can fail at a single CI run

PR #133, opened 2019-08-28
toolchain for android/firefox

Misaka-0x447f:

PR #134, opened 2019-08-29
toolchain for android/firefox

Artoria2e5:
was #133


SunriseFox:
+ we could just change the previou PR... without having to recreate it
+ web-ext could list android devices too (except that we need to analyse its output)... maybe reusing it could reduce our dependencies.

PR #140, opened 2019-08-30
chore(just): pointless rearrangement

Artoria2e5:


SunriseFox:
Today we made a new agreement, that only after you open an issue explaining what you want to do, and have objections settled, then shall you begin to make changes. Let us obey this rule.

Artoria2e5:
That's some really unnecessary thing done for a chore-level task with three lines of diff.

Jack-Works:
@sunrisefox pull request is also a kind of issue. (They share the same ID queue in the GitHub)

SunriseFox:
@Jack-Works No, we have to discuss before actually making changes.

PR #145, opened 2019-09-03
fix(editorconfig): you shall not put these rules there

Artoria2e5:
rules not under a file glob scope will simply do nothing

See also editorconfig/specification#8


Artoria2e5:
谁他妈把 master 的文件搞成过不了 prettier 了

Artoria2e5:
?!这不科学

PR #149, opened 2019-09-03
chore(linter): don't let jss escape prettier

Artoria2e5:


Jack-Works:
The JSS part is copy and mofidied from the JSS project. To following the upstream easier, we should not enforce too many things on it.

Artoria2e5:
they seem to be okay with the prettier checks, so I guess that's not too much of a concern?

PR #152, opened 2019-09-03
toolchain fixes

Misaka-0x447f:


Misaka-0x447f:
By tests, npm script manager actually has ability to kill
all process spawned by node even detached.
We recommended that do not running just tasks directly,
instead, use npm scripts.

Misaka-0x447f:
> 
> 
> I think, are we going in the wrong direction? Why should we care about how to kill a process group on *nix and Windows?

If not and not running scripts with npm script, a react-webpack dev server will remains in memory and hold the 3000 port until killed, this is a kind of memory leak.

SunriseFox:
Well... it becomes more and more complicated (and dangerous)... which should not have to be like this. Can we just remove `just` and get webpack to do these things? Something like `config-overrides-with-linter.js` or `config-overrides-with-firefox-(android).js`...

Misaka-0x447f:
This PR will not include chromium support because of upstream broken and not published release include this feature.

Misaka-0x447f:
@Artoria2e5 Why should we not break down big files to smaller parts by files and folders, but comment splitter? This could be hard to indexing. Anyway, you can keep your mind, I'm not strongly recommend to break it down since it exceeds 70 lines but not 1k lines.

Misaka-0x447f:
@Jack-Works You need to decide break down the config file or not.

SunriseFox:
have a look at #157

PR #157, opened 2019-09-04
chore: split webpack configuration

SunriseFox:
Motivation:
+ remove write-file-webpack-plugin
+ make it easier to add another plugin/loader
+ add the ability to extract more strings (e.g., banner text) to separate files. So we won't argue anymore about ``, '' or ""`

Actually done:
+ removed just*
+ moved firefox/android to webpack plugin, others to script
+ reworked eslint
+ added commit hook


SunriseFox:
Don't quite understand what happened.

```
Failed to compile.

./src/web-workers/QRCode.ts
Error: Debug Failure. Expected // ! TODO:
// ! find a easy way for using DOM context and worker context separeatly
// ! then we can remove @ts-ignore here and make worker type safe
import jsQR from 'jsqr';
// @ts-ignore
addEventListener('message', ev => {
    const [data, width, height] = ev.data;
    // @ts-ignore
    postMessage(jsQR(data, width, height));
});
 === undefined. Unexpected multiple outputs, file: .tscache/
```

Jack-Works:
> Don't quite understand what happened.
> 
> ```
> Failed to compile.
> 
> ./src/web-workers/QRCode.ts
> Error: Debug Failure. Expected // ! TODO:
> // ! find a easy way for using DOM context and worker context separeatly
> // ! then we can remove @ts-ignore here and make worker type safe
> import jsQR from 'jsqr';
> // @ts-ignore
> addEventListener('message', ev => {
>     const [data, width, height] = ev.data;
>     // @ts-ignore
>     postMessage(jsQR(data, width, height));
> });
>  === undefined. Unexpected multiple outputs, file: .tscache/
> ```
> 
> **Edit**: caused by building while serving

Just remove incremental build in tsconfig.

SunriseFox:
Preview: 
image
image

Misaka-0x447f:
image
Not working on my machine.

SunriseFox:
@Misaka-0x447f Did you run yarn install before that?

config-overrides/index.js
```diff
@@ -0,0 +1,159 @@
...
+            'We directly take the output to submit to the Web Store. We will integrate the automatic submission\n' +
+            'into the CircleCI in the near future.',
+        ),
+    )
+
+    // Write files to /public
+    if (env === 'development') {
+        config.plugins.push(
+            new (require('copy-webpack-plugin'))(
+                [...polyfills.map(from => ({ from, to: publicPolyfill })), { from: publicDir, to: dist }],
```
Misaka-0x447f: Sure. It said that polyfill is not defined, seems like you missed a import statement here.

SunriseFox: It should have been removed.

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "es2017",
     "lib": ["dom", "dom.iterable", "es2017", "esnext.asynciterable"],
-    "allowJs": false,
+    "allowJs": true,
```
Jack-Works: Why allowJS?

SunriseFox: so that js files in the `src` folder could be linted instead of just reporting an error... such as `setup.ssr.js` and future things...

PR #170, opened 2019-09-10
Chore/configuration

SunriseFox:
.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,12 @@
   },
   "scripts": {
     "start": "react-app-rewired start",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,9 +8,12 @@
   },
   "scripts": {
     "start": "react-app-rewired start",
+    "start:debugging": "node --inspect-brk node_modules/react-app-rewired/scripts/start.js",
```
Misaka-0x447f: It seems that @Artoria2e5 did more. Is it okay to cherry-pick this line to master?

PR #180, opened 2019-09-12
feat: conditional compilation for manifest

SunriseFox:


Jack-Works:
wait ill review later dont merge it now

SunriseFox:
I don't think 1 could be solved. And for 2, I think it is very likely that different code base will be used for different target in the future, so multiple build might still be ok. Or we could make some copies of the built artifact and put generated manifests in and pack.

Jack-Works:
1 can be resolved by just use many `manifest.json` files.

Different build target is good for eliminating codebase.

Propose: Use `process.env.platform` in the codebase to distinguish the following platforms
```ts
enum Platforms {
    Chrome, FirefoxDesktop, FirefoxMobile, GeckoView, WKWebKit
}
```

And when using development build, we manually set `process.env.platform` in our code to make a codebase that can run everywhere, when using production build on the `release/*` and `released` branch, we use a DefinePlugin to make it become a compile-time constant.

This can also unify how we detect platform in our codebase. (We have `isFirefox`, `isWKWebView`... now)

SunriseFox:
I don't quite know how IDE actually detect and use the schema. Can we tell it to follow our schema to lint specific files? Just as... no matter what type `.eslintrc` is, VSCode always knows about that.

Jack-Works:
VSCode for example. Plugins will tell JSON language service what scheme they want to use to lint the speical json file. Like TypeScript support for VSCode will tell language service that they want to lint `tsconfig.json` with scheme with http://json.schemastore.org/tsconfig (actually it's not. tsconfig's json schema file is served on Microsoft.).

For custom json file it is also possible to lint with schema, we use `"$schema": "http://json.schemastore.org/chrome-manifest"` to lint the manifest(Note: it's not a JSON standard). But we can't do this in a JS file. JS file is managed by TypeScript Language Service in VSCode and it will not parse JSON schema. (other editor may have their own way to do this.)

Jack-Works:
I think we can return to the old way, using many `manifest.json` and switch to conditional compilation, then fork a `maskbook-wkwebview` and `maskbook-geckoview`.

Therefore, wkwebview and geckoview will have their own release schedule, can choose to opt-in new feature branches and make pull request back when they have some stable fixes for their platforms.

What's your idea?
@SunriseFox @Misaka-0x447f @Artoria2e5 @guanbinrui 
cc @MainasuK @Tlaster

SunriseFox:
My opinion is that we split it to multi repos and handle them separately.

Artoria2e5:
Splitting that into multiple repos would mean a huge increase in our workload, and I am strongly against that. Conditional compilation alone would be enough to do the separation. If it worked for the cross-platform .NET apps, it should work for us.

Yes, I understand that `tsc` does not have that feature yet, but with some clever things we can effectively emulate every aspect of it. https://github.com/Microsoft/TypeScript/issues/449#issuecomment-521061571 mentions many tools that handle these things at the `webpack` level. There is also the technique of using multiple `tsconfig` files for different sub-projects and pointing `tsc` at them.

Jack-Works:
The core of the problem is not we need a conditional compilation. It is how we should treat Maskbook core codebase as a dependency for iOS and Android. It happened many times, that the main line goes too fast and always breaks the mobile side.

For forking repos, they can choose when to follow the main line. This can also done by make new branches for them but it's quite messy to have something like `master-ios` `master-android` `release-ios` `release-android`

Artoria2e5:
I would definitely prefer branches over permanent forks.

Jack-Works:
Okay whatever, we have cond compilation now. But it doesn't enabled by default since we doesn't have so much things that need cond compilation. We will have a runtime "process.env.target" if no build target is specified.

Jack-Works:
rebased to master

package.json
```diff
@@ -96,19 +109,12 @@
...
-      "not dead",
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
+    "production": [">0.2%", "not dead", "not op_mini all"],
+    "development": ["last 1 chrome version", "last 1 firefox version", "last 1 safari version"]
```
Artoria2e5: turn on prettier, please

PR #181, opened 2019-09-12
fix: bring copy polyfills back

SunriseFox:
This is based on t4, `feat: use manifest generator to generate manifests `


SunriseFox:
`adoptedStyleSheets.js` is directly referenced by `public/index.html`, even though it is require from `src\utils\jss\ConstructableStyleSheetsRenderer.ts`. Shall we remove the reference or copy it?

`polyfill/browser-polyfill.min.js` is directly referenced by all templates.

Jack-Works:
yes, just remove it, it's no longer used

PR #183, opened 2019-09-12
fix: drop shell for lint

Artoria2e5:
using `shell` breaks filenames with ampersands.

we can node it ourselves; triggered by nodejs/node#29532


scripts/lint.js
```diff
@@ -8,20 +8,30 @@ process.chdir(base)
...
-              ...(reportOnly ? ['--check', '--loglevel', 'log'] : ['--write', '--loglevel', 'warn']),
-          ])
-        : null
+    if (!argv.noEslint)
+        await spawn(
+            'node',
+            [
+                // Guess what happens if you don't do this?
+                // Path resolution makes a \n, which becomes...
+                path.join(base, 'node_modules/eslint/bin/eslint.js'),
```
Misaka-0x447f: This could be a problem in case of using pnpm.

Jack-Works: What about require.resolve('eslint') or something?

Artoria2e5: pnpm will not break because both are explicit dependencies.

Artoria2e5: doing a resolve sounds good but is pointless -- they are never going to resolve into global deps or the installation is broken. and if they are not global does they are going to be exactly there because they are explicitly listed as deps.

Jack-Works: reasonable. but hard code a path is little hack with me

Artoria2e5: aghflaljdgajdlaj

SunriseFox:
Now use eslint api to lint.

PR #178, opened 2019-09-12
fix(qr): QR scanning on firefox not working

SunriseFox:
# Changes:

Fix QR scanning on Firefox and #161 (by SunriseFox)

A dialog for requesting permissions (by SunriseFox and Jack):
image

Theme switched to indigo (400) and orange (800) (by Jack)

Switch to more standard Material (by Jack):
image

Better CJK fonts (by Jack):
image

PR #163, opened 2019-09-06
Auto share posts to friends (part 1)

Jack-Works:
# Changes
- UI: In PostBox, add all people to selection by default
- Database: Implements methods for UserGroup add/query/delete/modify

This will also fix #159 and #173

PR #189, opened 2019-09-18
Conditional compilation

Jack-Works:


SunriseFox:
It is mixing changes from different commits and branches. I think it might cause issues when merging to master. I suggest we should deal with #163 and #178 first or base our work on *t2* or *feature/auto-share-to-friends*.

Jack-Works:
Ready to merge back to `t4` @SunriseFox

PR #197, opened 2019-09-20
fix: organize polyfill testing in CSSR

Artoria2e5:


Artoria2e5:
我有空再整理一下尽量送给上游吧……

Jack-Works:
Our solution is not transparent to the normal developer and quite hacky. I think they may not receive our solution but it can be a useful reference about how to render in shadow roots

PR #199, opened 2019-09-20
fix: use system-ui font when possible

Artoria2e5:


Jack-Works:
Will test the display effect in CJK later

PR #209, opened 2019-09-23
new: android rework

SunriseFox:
+ remove unused popup.html and clear index.html (It's confusing that we have to have an index.html)
+ splice browser-polyfill from firefox
+ add browser.permissions polyfill for geckoview
+ copy polyfills even when dev rebuild


src/polyfill/permissions.js
```diff
@@ -0,0 +1,15 @@
+;(()=>{
+  if (!browser) return
+  if (browser.permissions) return
+  browser.permissions = {}
+  const requestedUrls = []
```
Jack-Works: You should save them to the localstorage or somewhere.

SunriseFox:
I think it is ok to not use set on such a small dataset, a set will add some extra overhead.
This script is only loaded from options-page, whenever the user opens it. So maybe not reloading on getAll is also ok.

Jack-Works:
> This script is only loaded from options-page, whenever the user opens it. So maybe not reloading on getAll is also ok.

When generate backups, it is called in the background script.

SunriseFox:
So... If `getAll` is called by background page, it means... no matter how I change this file here, it never affects the result being retrieved from the background page. Background page uses native ways to get and request permissions and this patch only applies to wherever there's not a `browser.permissions`.

Shall it replace the native `browser.permissions` anyway?

Jack-Works:
If you're patching it on the options page, it should also be patched in the background page

SunriseFox:
Should different artifact generated by different branches? I think you should settle it.

Jack-Works:
> Should different artifact generated by different branches? I think you should settle it.

Yes that will be nice. I'll do it afternoon.

PR #213, opened 2019-09-24
feature: conditional compilation

Jack-Works:


SunriseFox:
I suggest we only merge them to the master branch and re-sync the sub-branches with master.

Misaka-0x447f:
out of date, waiting for update before review.

config-overrides/manifest.overrides.js
```diff
@@ -9,18 +9,13 @@ function firefox(manifest) {
     manifest.permissions.push('tabs')
 }
 /**
+ * Geckoview is firefox with some quirks.
  * @param {typeof base} manifest
  */
-function firefoxGeckoview(manifest) {
```
Jack-Works: I'm afraid your changes are not right in some way. Merging Firefox and Firefox(Desktop and Android) is mixing the concept of Firefox as a browser engine and a platform. Maybe there are some bugs in Firefox browser engine itself so we have to add a field in the manifest, or maybe there are some bugs with Firefox Desktop/Android only. By merging them, we can't create a fix that for Firefox Desktop and Android only but not for geckoview.

Merging Firefox Desktop and Firefox Android because the Firefox Add-on store doen't allow us to upload seperate packages for them.

Jack-Works: @Artoria2e5

Artoria2e5: > mixing the concept of Firefox as a browser engine and a platform. 

No, my change is logically correct as it stands. It is saying that `geckoviewFuckery extends firefoxFuckery` and is doing its own special workarounds in addition to the `super` firefox workaround.

> maybe there are some bugs with Firefox Desktop/Android only

This is highly unlikely, given the way Firefox is going to proceed. Geckoview-specific modifications, on the other hand, should fade away in the future as the upstream plumbs through addon-related blockers and make it the new Firefox on Android.

Jack-Works: > Geckoview-specific modifications, on the other hand, should fade away in the future as the upstream plumbs through addon-related blockers and make it the new Firefox on Android.

We should not assume that since there are lots more for geckoview before it replace the offical Firefox

Artoria2e5: Assuming that geckoview will not have firefox desktop bugs is also a pretty big guess...

Jack-Works: Yeah so they are seperate before your commit. FirefoxDesktopAndMobile extends Firefox, GeckoView extends Firefox.

Artoria2e5: whatever, let's not die on this hill. rebase and remove that one commit.

Jack-Works:
@Artoria2e5 @SunriseFox @Misaka-0x447f @guanbinrui need review

PR #216, opened 2019-09-24
Feature/conditional compilation ci full

Jack-Works:


scripts/ci-build.js
```diff
@@ -0,0 +1,83 @@
+const { spawn } = require('./spawn')
+const path = require('path')
+
+const base = path.join(__dirname, '../')
+process.chdir(base)
+async function main() {
+    const prepareCommands = getCommands(`
+    yarn install --link-duplicates --frozen-lockfile
+    yarn lint:report
+    ${
```
Artoria2e5: ………………

Why the comment? And if you think that check is worth doing, you might want to ask why you are building at all…

scripts/ci-build.js
```diff
@@ -0,0 +1,83 @@
...
+        /**
+         * Let's do a quick TypeScript check.
+         */ 'yarn tsc -p tsconfig_cjs.json'
+    }
+    sudo apt-get install zip
+    `)
+    for (const [commands, ...args] of prepareCommands) {
+        await spawn(commands, args)
+    }
+    const currentBranch = (await getCurrentGITBranchName()).toLowerCase()
```
Artoria2e5: Git is not an acronym. "Git"./

scripts/ci-build.js
```diff
@@ -0,0 +1,83 @@
...
+    return platforms
+        .sort()
+        .map(generateCommand)
+        .join('\n')
+    /** @param {('base' | 'iOS' | 'chromium' | 'firefox' | 'gecko')} type */
+    function generateCommand(type) {
+        if (type === 'chromium' && platforms.indexOf('base') !== -1) {
+            // chromium doesn't have it's own changes yet.
+            // just copying base version is acceptable
+            return `
```
Artoria2e5: just return ""

scripts/ci-build.js
```diff
@@ -0,0 +1,83 @@
...
+            cp Maskbook.base.zip Maskbook.chromium.zip
+            `
+        }
+        return `
+        echo "Building for target ${type}"
+        yarn build:${type.toLowerCase()}
+        bash -c "cd build && zip -r ../Maskbook.${type}.zip ./* && cd .."
```
Artoria2e5: You don't need to `cd` back because it already exited

PR #217, opened 2019-09-24
Feature/conditional compilation ci full

Jack-Works:


config-overrides/manifest.overrides.js
```diff
@@ -24,7 +18,7 @@ function firefoxGeckoview(manifest) {
 /**
  * @param {typeof base} manifest
  */
-function firefoxDesktop(manifest) {
+function firefoxDesktopAndAndroid(manifest) {
```
Artoria2e5: Just remove this one and call it "firefox"

Artoria2e5:
然后我不知道你有没有发现，我们 cond compile 这个还没 merge 进去就开子 master branch 说明什么？说明这几个 branch 开早了。本来是可以 `checkout -b` 的事情。

config-overrides/manifest.overrides.js
Jack-Works: it's confusing because we have gecko which also is firefox

Artoria2e5: it's not hard, it's just a `geckoViewManifest extends firefoxManifest` concept

scripts/ci-build.js
```diff
@@ -52,20 +52,18 @@ function getBuildCommand(platforms) {
         if (type === 'chromium' && platforms.indexOf('base') !== -1) {
             // chromium doesn't have it's own changes yet.
             // just copying base version is acceptable
-            return `
-            cp Maskbook.base.zip Maskbook.chromium.zip
-            `
+            return ''
```
Artoria2e5: ……………………………………

Jack-Works:
> Just let it be, as we are already using them in the mobile branch.

Hmm Should I close this pr or ...?

Jack-Works:
> 然后那个 `return ''` 是搞什么鬼……我是让你写成一行的字符串不是让你搞成真的空字符串

It has the same effect when it applies to chrome

PR #253, opened 2019-10-09
Introduce new payload -38 to fix a critical hash problem

Jack-Works:
1. Introduce a new -38 payload
2. Apply the stable hash fix on the -38 version

```ts

    // ? payload is 🎼4/4|AESKeyEncrypted|iv|encryptedText|signature|publicShared:||
    // ? if publicShared is true, that means AESKeyEncrypted is shared with public
```

### DIFF
--- a/src/components/InjectedComponents/PostInspector.tsx
+++ b/src/components/InjectedComponents/PostInspector.tsx
@@ -57,7 +57,11 @@ export function PostInspector(props: PostInspectorProps) {
 
     if (type.encryptedPost) {
         if (!isDebugging) props.needZip()
-        const { iv, ownersAESKeyEncrypted, version } = type.encryptedPost
+        const { iv, version } = type.encryptedPost
+        const ownersAESKeyEncrypted =
+            type.encryptedPost.version === -38
+                ? type.encryptedPost.AESKeyEncrypted
+                : type.encryptedPost.ownersAESKeyEncrypted
         return (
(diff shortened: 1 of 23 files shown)


src/components/InjectedComponents/PostInspector.tsx
```diff
@@ -57,7 +57,11 @@ export function PostInspector(props: PostInspectorProps) {
 
     if (type.encryptedPost) {
         if (!isDebugging) props.needZip()
-        const { iv, ownersAESKeyEncrypted, version } = type.encryptedPost
+        const { iv, version } = type.encryptedPost
+        const ownersAESKeyEncrypted =
+            type.encryptedPost.version === -38
```
Misaka-0x447f: condition may need to be greater than or equal -38.

Misaka-0x447f:
We may need a payload version managing system just like the social-network one.

Tedko:
+1

Jack-Works:
> We may need a payload version managing system just like the social-network one, or the difficult of maintenance may increase quickly.

We do have one

Misaka-0x447f:
> 
> 
> > We may need a payload version managing system just like the social-network one, or the difficult of maintenance may increase quickly.
> 
> We do have one

We do not have one. I means that we need something like ```definePayloadVersion``` so that we will not need to care versioning outside of that file/folder.

PR #255, opened 2019-10-10
fix: firefox regression (bring by "chore: reduce the use of `any`")

Jack-Works:
This commit brings a regression on Firefox.

This reverts commit a5402a8456e1de0a882cbaf0560f6ce432cbb3f3.


Jack-Works:
related #258

PR #265, opened 2019-10-10
Sync master => master-android

Jack-Works:
This pr is making changes on the `master` branch into `master-android`

By approving this pull request, you agree to replace the current `master-android` branch with this branch.

This branch contains all changes from both `master` and `master-android`

WARN: Test it before approve this pull request on Android!


Misaka-0x447f:
Cannot run test on Android, dismissing review request

Jack-Works:
> The conflicts are too effed up; we should just approve it and use a force-push after we all agree.

Yes, I already do this in the ios branch

PR #282, opened 2019-10-15
feature: better UI on mobile device

Jack-Works:
Mobile device:

image

image

### DIFF
--- a/src/extension/background-script/WelcomeService.ts
+++ b/src/extension/background-script/WelcomeService.ts
@@ -228,3 +228,8 @@ export async function openWelcomePage(id?: SocialNetworkUI['lastRecognizedIdenti
     }
     return browser.tabs.create({ url: getWelcomePageURL(id) })
 }
+
+export async function openOptionsPage(url: string) {
+    if (url === '' || url === '/') url = 'index.html'
+    return browser.tabs.create({ url: browser.runtime.getURL(url) })
+}
--- a/src/extension/background-script/WelcomeService.ts
+++ b/src/extension/background-script/WelcomeService.ts
@@ -229,7 +229,6 @@ export async function openWelcomePage(id?: SocialNetworkUI['lastRecognizedIdenti
     return browser.tabs.create({ url: getWelcomePageURL(id) })
 }
 
-export async function openOptionsPage(url: string) {
-    if (url === '' || url === '/') url = 'index.html'
-    return browser.tabs.create({ url: browser.runtime.getURL(url) })
+export async function openOptionsPage(route: string) {
+    return browser.tabs.create({ url: browser.runtime.getURL('/index.html#' + route) })
 }
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -80,11 +80,8 @@ const useStyles = makeStyles(theme =>
             marginLeft: drawerWidth,
             zIndex: theme.zIndex.drawer + 1,
         },
-        menuButton: {
+        backButton: {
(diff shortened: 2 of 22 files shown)


src/extension/background-script/WelcomeService.ts
```diff
@@ -228,3 +228,8 @@ export async function openWelcomePage(id?: SocialNetworkUI['lastRecognizedIdenti
     }
     return browser.tabs.create({ url: getWelcomePageURL(id) })
 }
+
+export async function openOptionsPage(url: string) {
```
Artoria2e5: openExtensionPage?

src/index.tsx
```diff
@@ -167,39 +148,56 @@ function ResponsiveDrawer() {
...
                 </div>
             </Router>
         </ThemeProvider>
     )
 }
-function LinkItem(props: { to: string; icon: React.ReactElement; title: string }) {
+function MobileBackButton() {
+    if (!(webpackEnv.firefoxVariant === 'GeckoView' || webpackEnv.target === 'WKWebview')) return null
     return (
+        <MuiLink color="textPrimary" onClick={window.close}>
```
Artoria2e5: Is that a close button? Use a "close" button if your back button is really intended to close the popup, or use `window.history.back` for back. See the semantics at [material.io](https://material.io/design/navigation/understanding-navigation.html).

The spacing on that button looks awkward. Specifically:
* There should be more space to the right of the arrow. The placement should be identical to the `menuButton` element in index.ts.
* The baseline of the logo looks off. Should be fixed by applying the `menuButton` margin: as long as it's not directly next to the IconButton it should be fine.

src/social-network-provider/facebook.com/worker-provider.ts
```diff
@@ -9,6 +9,7 @@ export const facebookWorkerSelf = defineSocialNetworkWorker({
     ...sharedProvider,
     fetchPostContent: fetchPostContentFacebook,
     fetchProfile: fetchProfileFacebook,
-    autoVerifyBio: autoVerifyBioFacebook,
+    autoVerifyBio: webpackEnv.target === 'WKWebview' ? null : autoVerifyBioFacebook,
```
Artoria2e5: ……okay.

src/extension/background-script/WelcomeService.ts
Jack-Works: they are synonyms since we don't expect user to open background page or popup page

Artoria2e5:
#288, but it should be non-blocking.

src/index.tsx
Jack-Works: this button only appears on Android or iOS standalone app. there is no "tabs" equivalent in those environment so clicking on the back button behaves like "back" from the settings page to the Facebook.com

src/extension/background-script/WelcomeService.ts
Artoria2e5: `(url: string)`

Jack-Works:
> #288, but it should be non-blocking.

you can change from request change to approve

src/extension/background-script/WelcomeService.ts
Jack-Works: maybe `/index.html#/welcome` or something. ok I'll change that

Artoria2e5:
2em.

neruthes:
Please update UI according to https://github.com/DimensionDev/Maskbook_MobilePWA-Meta/issues/1

Issue #70, opened 2019-07-25
Dashboard Revamp


neruthes:
Components:

- Dashboard / Initialization
  - 
- Dashboard / Main
  - Add Persona
    - [PersonaCreation] Modal
      - Procedure:
        - Enter name
        - Get keypair
        - Connect first profile
  - Delete Persona
    - Popup Confirmation
  - Add Profile
    - [ProfileBinding] Modal
      - Procedure:
        - Enter profile URI
        - Get guides to paste public key into profile biography
  - Delete Profile
    - Popup Confirmation
  - Export Persona
  - Import Persona
  - Export Keystore
  - Import Keystore
- Dashboard / Synchronization
  - Add Device
  - Rename Current Device
  - Quit Current Misakanet

SunriseFox:
As we are going to support multiple identities and multiple providers, some changes we may need to apply to our dashboard:

+ Possibly rename current menu item `Setup your account` to `Initialize your Maskbook`, in which we could provide an all-in-one setup for all providers (facebook, twitter, etc.).
+ Add `Identity Management` menu item, allowing user to **add more keystores**, **select default keystore for each provider**, **disable/enable each keystore** (e.g. if enabled, we may try using this keystore to decrypt otherwise not), **retrieve public keys** (create/recreate proof post, #79 ) or completely **remove them**. 

As a user may use more than one identity and thus own multiple keystores, cross-device transfer by a QR code which contains full keystore information seems not quite practical. So maybe a (distributed) bridge server is still needed. #77 we create a secure channel, and sync all / user-selected keystore by this channel with trusted pair.

Before that we also have to decide how the user may identify their identities, or rather, keystores.  A user with multiple keys may get confused when managing them. We may assign a randomly generated English word to each keystore, or let user 'tag' it manually. Any other better solutions? 

##### Summary in this issue

+ New initialization page for more providers
+ New setting page for identity management
+ New approach for tagging identities

Tedko:
<img width="338" alt="Screen Shot 2019-08-22 at 00 39 27" src="https://user-images.githubusercontent.com/6687437/63496061-07597900-c476-11e9-8c11-9be21b496058.png">
<img width="334" alt="Screen Shot 2019-08-22 at 00 39 52" src="https://user-images.githubusercontent.com/6687437/63496064-09bbd300-c476-11e9-8ca1-e43f885d1874.png">


Mentioned in #79

neruthes:
Preview.

[...]

PR #284, opened 2019-10-15
introduce the mobile version close #232

### LINKED ISSUE #232: Mobile QR is not efficient enough
https://github.com/DimensionDev/Maskbook/issues/186#issuecomment-533059735:
> When there are more than 1 identities the QR code becomes too large to scan. So we really need to find another way to sync between devices.

Our current QR encoding scheme is woefully inefficient: we put a json with long key names and base64-ed data through QR's 8-bit "byte" encoding, and get a huge image back. 

### The JSON level

The problems with `JsonWebKey` noted here should be handled by #224.

<details>
<summary>A big ol' backup (636b minimized)</summary>

```JSON
{
    "version": 1,
    "whoami": [
        {
            "network": "facebook.com",
            "userId": "100034588593645",
            "previousIdentifiers": [],
            "localKey": {
                "alg": "A256GCM",
                "ext": true,
                "k": "REDACTEDREDACTEDREDACTEDREDACTEDREDACTED***",
                "key_ops": [
                    "encrypt",
                    "decrypt"
                ],
                "kty": "oct"
            },
            "publicKey": {
                "crv": "K-256",
                "ext": true,
                "x": "7seqxT72nd8UhC8tIWKqux0LWT-de7sDVSYqO4vLJnI",
                "y": "oNRjB320h3p4TUbjZmHk5iZByPp0Q6aT4MX9BpA2mmM",
                "key_ops": [
                    "deriveKey"
                ],
                "kty": "EC"
            },
            "privateKey": {
                "crv": "K-256",
                "ext": true,
                "d": "REDACTEDREDACTEDREDACTEDREDACTEDREDACTED***",
                "x": "7seqxT72nd8UhC8tIWKqux0LWT-de7sDVSYqO4vLJnI",
                "y": "oNRjB320h3p4TUbjZmHk5iZByPp0Q6aT4MX9BpA2mmM",
                "key_ops": [
                    "deriveKey"
                ],
                "kty": "EC"
            }
        }
    ]
}
```
</details>

What problems are we seeing here?

* We are saving both the public and the private key when they have the exact same `x` and `y`. Of course they do; they are the same effing key.
 * The maths say that we can get `y` from `x`, and we actually use that for importing the key.
 * This ain't the worst part: the maths also say we can get the public `x` and `y` from the private `d`. We are needlessly wasting five times the space for all these.

Let's yank down the useless parts and try again. Boom! 364b minimized.

### Other corners
There are some small overheads that can be reduced by:
* JSON: binary formats. Pick one, pack it.
* Base64: compression. Using json already forces us onto the less-efficient byte (as compared to `alphanumerical`) encoding of QR with the special characters it uses, so why don't we push it further? Even `zlib` can effectively revert the damage base64 does to the data size.

Jack-Works:
image

Must merge after
- Gun upgrade #271
- Better UI on mobile #282  

close #232

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ For general introductions, see https://Maskbook.com/
     [crext]: https://chrome.google.com/webstore/detail/maskbook/jkoeaghipilijlahjplgbfiocjhldnap/
     [fxaddon]: https://addons.mozilla.org/en-US/firefox/addon/maskbook/
     [ios]: https://testflight.apple.com/join/OGmGmIg1
-    [android]: https://play.google.com/apps/testing/com.dimension.maskbook
+    [android]: https://play.google.com/store/apps/details?id=com.dimension.maskbook
 
 ## Documentation for developers
 
--- a/README.md
+++ b/README.md
@@ -5,10 +5,14 @@ Encrypt your posts & chats on You-Know-Where. Allow only your friends to decrypt
 For general introductions, see https://Maskbook.com/
 
 -   [Maskbook on Chrome Web Store][crext]
--   [Maskbook on Firefox Addon Store][fxaddon] (Works on Android too!)
+-   [Maskbook on Firefox Addon Store][fxaddon] (Works on Firefox for Android too!)
+-   (Beta!) [Maskbook for iOS][ios]
+-   (Beta!) [Maskbook for Android][android]
 
     [crext]: https://chrome.google.com/webstore/detail/maskbook/jkoeaghipilijlahjplgbfiocjhldnap/
     [fxaddon]: https://addons.mozilla.org/en-US/firefox/addon/maskbook/
+    [ios]: https://testflight.apple.com/join/OGmGmIg1
+    [android]: https://play.google.com/apps/testing/com.dimension.maskbook
 
 ## Documentation for developers
 
--- a/src/components/MobileImportExport/Export.tsx
+++ b/src/components/MobileImportExport/Export.tsx
@@ -62,7 +62,7 @@ export function ExportData() {
                     iOS
                 </Link>
                 {' and '}
-                <Link target="_blank" href="https://play.google.com/apps/testing/com.dimension.maskbook">
+                <Link target="_blank" href="https://play.google.com/store/apps/details?id=com.dimension.maskbook">
                     Android
                 </Link>
                 {' version'}
--- a/src/components/MobileImportExport/Export.tsx
+++ b/src/components/MobileImportExport/Export.tsx
@@ -2,23 +2,26 @@ import React, { useState } from 'react'
 import { QrCode } from '../shared/qrcode'
 import { useAsync } from '../../utils/components/AsyncComponent'
 import Services from '../../extension/service'
-import { makeStyles, Typography } from '@material-ui/core'
+import { makeStyles, Typography, Button } from '@material-ui/core'
 import { NotSetupYetPrompt } from '../shared/NotSetupYetPrompt'
 import { geti18nString } from '../../utils/i18n'
 import { useCurrentIdentity, useMyIdentities } from '../DataSource/useActivatedUI'
 import { ChooseIdentity } from '../shared/ChooseIdentity'
+import { compressBackupFile } from '../../utils/type-transform/BackupFileShortRepresentation'
+import { BackupJSONFileLatest } from '../../utils/type-transform/BackupFile'
 
 const useStyles = makeStyles({
     root: {
         textAlign: 'center',
         paddingTop: 24,
+        maxWidth: 600,
+        margin: 'auto',
         '& > div': {
             margin: 'auto',
         },
     },
     code: {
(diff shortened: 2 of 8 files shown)


src/components/MobileImportExport/Export.tsx
```diff
@@ -44,6 +53,26 @@ export function ExportData() {
     }
     return (
         <main className={classes.root}>
+            <Typography style={{ marginBottom: 24 }}>
+                <Button style={{ marginRight: 12 }} size="small" variant="outlined" color="primary" disabled>
+                    Beta
+                </Button>
+                {'Checkout Maskbook '}
```
Artoria2e5: "Check out"

src/components/MobileImportExport/Export.tsx
```diff
@@ -44,6 +53,26 @@ export function ExportData() {
...
+                <Link target="_blank" href="https://testflight.apple.com/join/OGmGmIg1">
+                    iOS
+                </Link>
+                {' and '}
```
Artoria2e5: Too many separate strings here. I suggest that we use a `<Fragment>` React element and put direct HTML in here instead. That will make l10n easier too, as not all languages express this meaning in this exact order of words.

src/components/MobileImportExport/Export.tsx
```diff
@@ -44,6 +53,26 @@ export function ExportData() {
...
+                <Link target="_blank" href="https://play.google.com/apps/testing/com.dimension.maskbook">
+                    Android
+                </Link>
+                {' version'}
```
Artoria2e5: versions

src/components/MobileImportExport/Export.tsx
```diff
@@ -44,6 +53,26 @@ export function ExportData() {
...
+            </Typography>
+            <Typography style={{ marginBottom: 24 }}>
+                {'Maskbook is also available on '}
```
Artoria2e5: Same Fragment suggestion as above.

src/components/Welcomes/QRScanner/index.tsx
```diff
@@ -17,7 +18,7 @@ export default function QRScanner(
     const { scanning, onResult, onError, ...videoProps } = props
     const video = useRef<HTMLVideoElement | null>(null)
 
-    useQRCodeScan(video, scanning, onResult, onError)
+    useQRCodeScan(video, scanning, x => compressBackupFile(decompressBackupFile(x)), onError)
```
Artoria2e5: Feels like we are breaking some abstraction here. If you are ditching onResult, remove it completely.

Artoria2e5: … and what is this meant to do exactly? To make sure that the key is a valid compressed point?

README.md
```diff
@@ -5,10 +5,14 @@ Encrypt your posts & chats on You-Know-Where. Allow only your friends to decrypt
...
 -   [Maskbook on Chrome Web Store][crext]
--   [Maskbook on Firefox Addon Store][fxaddon] (Works on Android too!)
+-   [Maskbook on Firefox Addon Store][fxaddon] (Works on Firefox for Android too!)
+-   (Beta!) [Maskbook for iOS][ios]
+-   (Beta!) [Maskbook for Android][android]
 
     [crext]: https://chrome.google.com/webstore/detail/maskbook/jkoeaghipilijlahjplgbfiocjhldnap/
     [fxaddon]: https://addons.mozilla.org/en-US/firefox/addon/maskbook/
+    [ios]: https://testflight.apple.com/join/OGmGmIg1
+    [android]: https://play.google.com/apps/testing/com.dimension.maskbook
```
Tlaster: Or this `https://play.google.com/store/apps/details?id=com.dimension.maskbook`?

Tlaster: Or this `https://play.google.com/store/apps/details?id=com.dimension.maskbook`?

src/components/Welcomes/QRScanner/index.tsx
Jack-Works: Oh it's a mistake let me fix it.

src/components/MobileImportExport/Export.tsx
Jack-Works: It's a temporary UI. Can't we leave it as the current state?

Artoria2e5: I am not supposed to rule on that. Ask the PM or designer.

Jack-Works: @neruthes

neruthes: Since it is just a rush demo, I would prefer to conclude things in #70 instead of making issue-specific changes over the current jam.

neruthes: I feel no need to rush to refactor i18n for now. If you find changing i18n framework after delivering #70 will be greatly harder, we may further investigate. Current priority is #287 and #70.

Issue #176, opened 2019-09-11
Assign 172.104.123.119 to {gun,matrix}.maskbook.io/com domain


Jack-Works:

neruthes:
@asonmine Hope you may have a look into this matter.

asonmine:
Assigned.

asonmine:
Assigned all.

Tedko:
ok we may close this after test.

PR #285, opened 2019-10-15
chore: change domain to gun.maskbook.io related [#176]

Jack-Works:


Jack-Works:
> What about getting a SSL cert and switch to HTTPS...?

@neruthes @Tedko Should we?

cc @asonmine @yisiliu

neruthes:
Just wondering why `.io` instead of `.com` 🤔

neruthes:
I think CloudFlare can be enough in most cases.

SunriseFox:
But Cloudflare won't proxy any ports other than 80/443... If we use it, we need to switch to 443 port (recommended).

Jack-Works:
> Just wondering why `.io` instead of `.com` 🤔

Don't you think `.io` is cooler? 🤔

> But Cloudflare won't proxy any ports other than 80/443... If we use it, we need to switch to 443 port (recommended).

It's okay to use any port for gun

neruthes:
If GUN has an HTTPS interface, I suppose that CloudFlare is the most desirable option for now.

As for domain name, I do not find using `.io` necessary or fancy in this particular usage. Maybe I will find it cool in some other usage.

SunriseFox:
- [x] move maskbook.io / maskbook.com's NS to Cloudflare
- [ ] configure nginx with Cloudflare's origin ssl cert on GUN's server, proxy pass to localhost:8765
- [ ] configure Cloudflare to use Cloudflare DNS for `gun.maskbook.io/.com`
- [ ] merge this

SunriseFox:
if @yisiliu approved this idea, I should be able to deal with the server-side chores. Maybe [https://github.com/Dletta/visualGraph](https://github.com/Dletta/visualGraph) could be deployed too.

yisiliu:
Let's do this.

PR #291, opened 2019-10-16
fix: multiple identity support

SunriseFox:


.eslintrc.json
```diff
@@ -8,6 +8,7 @@
         "node": true
     },
     "rules": {
+        "no-debugger": "error",
```
Jack-Works: 🤔

src/extension/background-script/CryptoServices/decryptFrom.ts
```diff
@@ -138,7 +138,7 @@ export async function* decryptFromMessageWithProgress(
         }
 
         try {
-            if (by.equals(whoAmI)) {
+            try {
```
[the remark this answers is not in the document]
SunriseFox: move this try block outside is ok... but put it inside... seems also ok

.eslintrc.json
SunriseFox: in case I commit with `debugger` statement again

src/components/shared-settings/settings.ts
```diff
@@ -18,3 +21,28 @@ export const disableOpenNewTabInBackgroundSettings = createNewSettings<boolean>(
             "Many of Maskbook features relies on this behavior. Disable this behavior will limit Maskbook's functionality",
     },
 )
+
+const createProxiedSettings = (key: string) => {
+    const target: { [key: string]: ValueRef<string> | string } = {}
```
Jack-Works: Should this `target` defined in the global scope?

SunriseFox: no. it is only for current settings.

PR #317, opened 2019-10-22
feat: Allow group as a share target & remove deprecates

Jack-Works:
- now group can be the share target
- a better mech for showing myself's nickname
- change the signing process for the payload (might be breaking)
- drop migration file for old database (deprecated at Jun 1)
- upgrade post database to version 3 for future
- now maskbook will not be able to append recipients for dangerous version -40
- maskbook will store the aes key into the database when compose a new post

This part does not include any process about automatically share the post to someone.

image


neruthes:
Nice to see it being introduced ([I:b])

Issue #327, opened 2019-10-28
Add a Maskbook tip for public key on bio 


Tedko:
As the title.

For now we only have hint/tip for proof post.

SunriseFox:
Preview: 

Desktop:
![image](https://user-images.githubusercontent.com/14029711/68008932-6c71ec00-fcbb-11e9-8198-a39d0c8d83e6.png)

Mobile:
![image](https://user-images.githubusercontent.com/14029711/68008983-93c8b900-fcbb-11e9-9be8-6deee453cab4.png)

However it is not checking the bio but checking our database if it HAS been stored. This hint won't show if you just view someone's profile without any encrypted post observed.

Tedko:
> Preview:
> 
> Desktop:
> ![image](https://user-images.githubusercontent.com/14029711/68008932-6c71ec00-fcbb-11e9-8198-a39d0c8d83e6.png)
> 
> Mobile:
> ![image](https://user-images.githubusercontent.com/14029711/68008983-93c8b900-fcbb-11e9-9be8-6deee453cab4.png)
> 
> However it is not checking the bio but checking our database if it HAS been stored. This hint won't show if you just view someone's profile without any encrypted post observed.

Great!

neruthes:
Nice

PR #352, opened 2019-11-01
feat: add hint in bio (facebook only)

SunriseFox:
#327, this implementation reused `resolveLastRecognizedIdentityFacebook.ts`. At least there is one...

Desktop:
image

Mobile:
image

It is not checking the bio but checking our database if it HAS been stored. This hint won't show if you just view someone's profile without any encrypted post observed.


Jack-Works:
Should create a new function in the Social Network UI. Like inspectProfile()

Tedko:
> Should create a new function in the Social Network UI. Like inspectProfile()

+1

SunriseFox:
I'm not quite good at creating a new function...

Tedko:
Hi Jack

SunriseFox:
This shall be solved in #364.

PR #364, opened 2019-11-05
sync between branches & new features

SunriseFox:
+ Sync codes between master & master-ios

+ App bar

> image

+ Icon and overflow more mobile-friendly

> image

+ Hint word break opportunity to prevent overflow

> image

+ injectOptionsPageLink use same style of facebook

> image

+ Hint in bio

> Desktop:
> image
> 
> Mobile:
> image

+ improved 1000x performance (

+ Progress 

> image

Close SujiTech/Maskbook-iOS#2.
Close SujiTech/Maskbook-iOS#9.


SunriseFox:
Concern:

+ image

+ image

+ Performance is very very bad

src/index.tsx
```diff
@@ -172,8 +178,20 @@ function Dashboard() {
             .then(() => setExportLoading(false))
     }
 
+    const shouldRenderBackButton = useMediaQuery('(max-width:600px)')
```
Jack-Works: 🤔What about webpackEnv?

src/index.tsx
```diff
@@ -172,8 +178,20 @@ function Dashboard() {
...
+
     return (
         <Router>
+            <AppBar position="sticky">
+                <Toolbar>
+                    {shouldRenderBackButton && (
+                        <IconButton onClick={() => window.close()} edge="start" color="inherit" aria-label="menu">
```
Jack-Works: Wrong aria-label is worse than no aria-label

src/social-network-provider/facebook.com/UI/injectKnownIdentity.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export function injectKnownIdentityAtFacebook(this: SocialNetworkUI) {
+    let umount: () => void = () => {}
+    const self = othersBioLiveSelectorMobile.clone().concat(othersBioLiveSelectorPC)
+
+    const watcher = new MutationObserverWatcher(self)
+        .setDOMProxyOption({
+            afterShadowRootInit: { mode: 'closed' },
+        })
+        .addListener('onAdd', () => {
```
Jack-Works: Why not use `useForeach`? 🤔

src/index.tsx
SunriseFox: There's no choice about `mobile` in webpackEnv...

src/index.tsx
```diff
@@ -185,7 +185,7 @@ function Dashboard() {
             <AppBar position="sticky">
                 <Toolbar>
                     {shouldRenderBackButton && (
-                        <IconButton onClick={() => window.close()} edge="start" color="inherit" aria-label="menu">
+                        <IconButton onClick={() => window.close()} edge="start" color="inherit" aria-label="back">
```
Jack-Works: (actually, aria-label will be read out by the screen reader. if you want to use a aria-label, you also need to do the i18n here.)

src/index.tsx
Jack-Works: If you use mediaquery here, and someone use it on a Android tablet or iPad, they will find they have no way to exit the screen because width is bigger than 600px

SunriseFox: :awsl: so let's add one def to indicate if it is built for desktop...

src/utils/jss/ConstructableStyleSheetsRenderer.ts
```diff
@@ -41,19 +41,18 @@ function getStyleSheet(x: HTMLStyleElement) {
...
+        const styles = Array.from(fakeHead.children).filter(
+            (x): x is HTMLStyleElement => x instanceof HTMLStyleElement && x.innerText !== '',
+        )
         const shadows = Array.from(livingShadowRoots)
         const nextAdoptedStyleSheets = styles.map(getStyleSheet)
         for (const shadow of shadows) {
-            // Maybe we can remove this now.
             try {
-                shadow.adoptedStyleSheets = nextAdoptedStyleSheets
+                if (shadow.adoptedStyleSheets.length !== nextAdoptedStyleSheets.length) {
```
Jack-Works: Comparing the length may bring inconsistency in the styling. Changes on CSSStyleSheets (not only add or remove) will also invoke this function too

src/index.tsx
Jack-Works: !geckoView && !WKWebKit is enough

src/utils/jss/ConstructableStyleSheetsRenderer.ts
SunriseFox: Ah... I have another idea... can we merge these styles into one element... to improve performance?

Jack-Works: CSSStyleSheet object has internal state. Some methods depends on the order of the CSS rule. If you merge them, you have to maintain the order relationship.

Theoretically check .innerText also brings inconsistency. You should also check if there is any "rules" in the object (the innerText does not sync with the internal "rules")

Jack-Works:
need test & approve from @MainasuK

MainasuK:
> need test & approve from @MainasuK

Which commit?

Jack-Works:
> > need test & approve from @MainasuK
> 
> Which commit?

The latest one on this branch

MainasuK:
The app may crash due to this infinite RPC call:

1. Tap "Maskbook Settings" in the settings panel.
2. Tap "Create Backup"
3. Console print Infinite RPC log

MainasuK:
Can't tell. I did not set the benchmark and hard to set.

Jack-Works:
@SunriseFox maybe you need to rebase again 👀

Artoria2e5:
GitHub 这个 PR 的 merge 是 master onto ios，不是 ios onto master。如果大家都同意的话建议手动把 ios 按上 master 然后关了。

Jack-Works:
> GitHub 这个 PR 的 merge 是 master onto ios，不是 ios onto master。如果大家都同意的话建议手动把 ios 按上 master 强推然后关了。

这个 PR 是合并 master 和 master-iOS。Approve 之后会直接用这个 branch 替换 master-iOS 然后 merge 进 master

PR #393, opened 2019-11-13
feat: impl new partition on gun db

Jack-Works:
See spec change: https://github.com/DimensionDev/Maskbook/wiki/Data-structure-on-Gun-version-2/_compare/95bbca01b96285fc01d0d655c89a9e768c7e191f...7e34c815b33e52848fa26348b202cd09b8bbd61c


neruthes:
I suggest that the prefix should range in a polymorphism of the domain names of these social network platforms. For example, `twitter.com:` may map to `twitter--com---`, in good faith that no reasonable website will use consecutive hyphens except <https://git--help.com/>.

Also, I acknowledge that denying to use a simple `twitter.com:` is a direct result of a bug in GUN, so it may be good to investigate whether the bug remains.

Artoria2e5:
>  is a direct result of a bug in GUN

In other words, `gun.get('example.com:fooBar')` breaks stuff? Sounds like something to tell mark about...

neruthes:
> > is a direct result of a bug in GUN
> 
> In other words, `gun.get('example.com:fooBar')` breaks stuff? Sounds like something to tell mark about...

Reportedly fixed; but I did not in person witness such a commit message which states so from Mark.

* * *

Probably @amark may have some story to tell? ([I:b])

amark:
@neruthes @Jack-Works @Artoria2e5 this I think got fixed in the `0.9.x` to `0.2019.x` change as we had to change how RAD saved data (which broke old data storage) in order to fix the `.` bug. Should have been a dumb and obvious mistake I didn't make to begin with, sorry about that.

Issue #409, opened 2019-11-19
[Demand] Network-Specific Guides Revamp


neruthes:
## Metadata

Field | Value
--- | ---
Previous Stage | ...
Figma Link | [figma.com/...](https://www.figma.com/file/oFrOUZmDnOWL2pu6SNf4P7/Dashboard-v2?node-id=196%3A972)
Expected UX prototype due | 2019-11-22
Actual UX prototype release | 2019-11-21
Expected engineering due | 2019-12-29

### Current Stage

- Entry
- MBGuideBar Module
- Components:
  - Guides on Facebook
  - Guides on Twitter

### Left For Future Stages

- Backup Alert

## Entry

<img width="643" alt="image" src="https://user-images.githubusercontent.com/2139261/70430003-b9db3900-1a71-11ea-9553-e0979a6c4824.png">

If the user specifies his username and clicks "Next" button, the user will be directed to a 5-step guide.

If the user clicks this "I do not know my username..." button, the user will be direct to a guide which is a little different in terms of steps.

## MBGuideBar Module

<img width="1394" alt="image" src="https://user-images.githubusercontent.com/2139261/69314756-d59aae80-0c2c-11ea-8df4-2b69bc819be9.png">

Mainly designed for wide screens. For smaller screens, minor adjustments may be necessary.

### Structure

- Header
- Content
  - Title
  - Body (usually 2 lines)
- Buttons Area
  - Right Button (Next Step / Finish)
  - Left Button (Prev Step)
- Close Button

### Connect Profile

Title: `Current task: Connect profile @%s on facebook.com`

Step | Body Line 1 | Body Line 2 | Right Button | Left Button | Close Button
--- | --- | --- | --- | --- | ---
1 | Step 1/5: Login | Login on facebook.com as @%s | Next Step | - | End Guide
2 | Step 2/5: Navigate to your profile | Visit <a>https://www.facebook.com/%s</a>. | Next Step | Prev Step | End Guide
3 | Step 3/5: Enter bio editing mode | Click button "Edit bio". | Next Step | Prev Step | End Guide
4 | Step 4/5: Add public key to profile bio | <a>Click here</a> to copy public key. Paste it to your profile bio, and save profile. | Next Step | Prev Step | End Guide
5 | Step 5/5: Verification | Please wait as Maskbook is trying to check whether the setup is successful... | Finish (%ns) (disabled) (%n is a countdown from 8) | Prev Step | End Guide
[...]

Issue #407, opened 2019-11-19
[Demand] Initialization Revamp


neruthes:
## Metadata

Field | Value
--- | ---
Previous Stage | ...
Figma Link | [figma.com/...](https://www.figma.com/file/oFrOUZmDnOWL2pu6SNf4P7/Dashboard-v2?node-id=134%3A2)
Expected UX prototype due | 2019-11-22
Actual UX prototype release | 2019-11-21
Expected engineering due | 2019-12-15

## Catalogue

- Dashboard Logics
- Components:
  - Dashboard Pre-Initialization State
  - Dashboard Half-Initialized State

## Dashboard Logics

Dashboard has 3 states:

State Name | Condition
--- | ---
Pre-initialization state | 0 persona in database
Half-initialized state | 1 persona and 0 profile in database
Post-initialization state | otherwise

If there is no Persona currently, Dashboard will be in pre-initialization state; otherwise, Dashboard will be in post-initialization state. This selection is made every time when Dashboard is loaded.

Post-initialization state has been implemented in #369. 

## Component: Dashboard: Pre-Initialization State

In this state, Dashboard introduces Maskbook to a new user.

<img width="918" alt="image" src="https://user-images.githubusercontent.com/2139261/69315961-b0f40600-0c2f-11ea-9e7e-5621eecd80c6.png">

By entering `S.a.2` screen after the name of the new persona is given, now this Maskbook installation comes into half-init state.

### New User

<img width="1384" alt="image" src="https://user-images.githubusercontent.com/2139261/69318696-a50b4280-0c35-11ea-81f3-0524ffa2c861.png">

We onboard new users like this.

When clicking in-cell "Connect facebook.com" text button, the user will be prompted a [dialogue](https://www.figma.com/file/oFrOUZmDnOWL2pu6SNf4P7/Dashboard-v2?node-id=107%3A0), then the user will be directed to the profile page with a Guide Bar (#409) displayed on the top of the web page.

If we deliver this without #409, in the meantime, the [original behavior](https://www.figma.com/file/oFrOUZmDnOWL2pu6SNf4P7/Dashboard-v2?node-id=107%3A36) as defined in #369 will still be in use till the delivery of #409. 

### Returning User

<img width="949" alt="image" src="https://user-images.githubusercontent.com/2139261/69532113-0bad9a80-0f6d-11ea-89f3-432858fb0a4a.png">

[...]

Issue #369, opened 2019-11-08
Dashboard Revamp (Stage 2)


neruthes:
## Metadata

Field | Value
--- | ---
Previous Stage | #70 
Figma Link | [figma.com/...](https://www.figma.com/file/oFrOUZmDnOWL2pu6SNf4P7/Dashboard-v2?node-id=1%3A2)
Expected UX prototype due | 2019-11-12
Actual UX prototype release | 2019-11-13
Expected engineering due | 2019-12-15

## Catalogue

- Dashboard internal navigation structure minor adjustment
- Create Persona UX adjustment
- Connect Profile UX adjustment
- Delete Persona UX adjustment
- Disconnect Profile UX adjustment

### Left For Future Stages

- Database Backup
- Backup Persona UX adjustment
- Import Persona UX adjustment
- Debug Tool UX adjustment
- Misakanet UX initialization

### Desktop Browser Extension

Generally, Dashboard is opened in a new tab. In such scenario, we do not need a way to exit Dashboard.

### Stand-Alone App, Big Screen

Min-width: 1024.

In addition to the case of Desktop Browser Extension, we need a way to exit Dashboard. Simply add an AppBar above at the top left position, within the pseudo-drawer.

### Stand-Alone App, Small Screen

Max-width: 1023.

In this case, we need an AppBar which has a close button at the top-right corner for exiting. Also, in the sense that we cannot use a drawer button to invoke the drawer, the items in the drawer will be shown as tabs below the AppBar.

## Dashboard Home

- Multiple sections:
  - My Personas
  - Add Persona
  - Database

<img width="336" alt="Screen Shot 2019-11-12 at 10 09 31" src="https://user-images.githubusercontent.com/2139261/68662361-908fc180-0534-11ea-95d0-541c1ab02638.png">

##### Component: Profile Row

- Vacant Mode
  - Show a blue text button to connect a profile
- Filled Mode
  - Show a black text label to display the username with a leader `@` regardless of site-specific conventional representation of usernames.

##### Modes

- Standard Mode
  - Hover Persona Card to show "Edit Name" button
  - Hover Profile Row to show "Disconnect" button
- Name Edit Mode
  - "Save" button never hides
  - Other operations do not interrupt name edit mode

## Create Persona

- Step 1: Set name
- Step 2: See success screen, be suggested to connect a profile
- Step 3a: Quit
- Step 3b: Connect a profile

[...]

PR #441, opened 2019-12-02
feat: add verification bio on create or restore

SunriseFox:
Something here...
image

And here:
image

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -108,8 +108,11 @@
     "start": {
         "message": "Start"
     },
-    "verify": {
-        "message": "Verify"
+    "retry": {
+        "message": "Retry"
+    },
+    "verifying": {
+        "message": "Verifying..."
     },
     "skip": {
         "message": "Skip"
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -105,6 +105,15 @@
     "finish": {
         "message": "Finish"
     },
+    "start": {
+        "message": "Start"
+    },
+    "verify": {
+        "message": "Verify"
+    },
+    "skip": {
+        "message": "Skip"
+    },
     "undo": {
         "message": "Undo"
     },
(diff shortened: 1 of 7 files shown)


Jack-Works:
lgtm but what about @neruthes ?

neruthes:
Seems that these will be removed when #369 and #407 delivered? I would refrain from encouraging it unless there is a good chance that these will remain useful after #369 and #407. 

Specifically, one more step of verification in the procedure of connecting profile can be ok, while the verification upon database restoration looks confusing to me. The user should have a chance to update profile bio after successful restoration, instead of being rejected when trying to restore.

Jack-Works:
> Seems that these will be removed when #369 and #407 delivered? I would refrain from encouraging it unless there is a good chance that these will remain useful after #369 and #407.

we still need a bio prove after all

neruthes:
> > Seems that these will be removed when #369 and #407 delivered? I would refrain from encouraging it unless there is a good chance that these will remain useful after #369 and #407.
> 
> we still need a bio prove after all

I believe that the test-and-verify is well included in #369 and a more intuitive profile connecting guides wizard is well documented in #409. Let me know if there is any opinion regarding the coverage of #369 and #409.

neruthes:
## Request for changes

### State 1 (Base)

image

Looks good. No changes needed.

### State 2 (Ready)

Screen Shot 2019-12-04 at 08 55 27

2 buttons:

Position | Color | Text
Left | White | Retry
Right | Blue | Next

### State 3 (Verifying)

Screen Shot 2019-12-04 at 08 55 23

2 buttons:

Position | Color | Text
Left | White | Retry
Right | Blue (Disabled) | Verifying...

SunriseFox:
Should have done with these changes. You may try again.

Jack-Works:
how does the ui works on the mobile? it might be not so easy to switch between tabs

Jack-Works:
I'm wondering if we can replace the technical word "verify" with "Check if I have set Maskbook up yet", that may reduce the confusion to users

neruthes:
> I'm wondering if we can replace the technical word "verify" with "Check if I have set Maskbook up yet", that may reduce the confusion to users

Too long to put into a button. 🤔

Jack-Works:
We have a new kind of verification on the bio page, does that enough? (Maybe not, not every one will check their profile page frequently)

neruthes:
After #369 is delivered, the profile connecting procedure requires successful setup in the profile bio. #409 will further strengthen this.

PR #513, opened 2019-12-12
perf: let gun1 share instance with gun2

Jack-Works:


Jack-Works:
@SunriseFox any progress on this?

Jack-Works:
@sunrisefox hmm did you add the new gun server?

SunriseFox:
I don't know if I should... No more directions since deployment.

Tedko:
Better add this new gun instance — Did you test it?

Jack-Works:
> Better add this new gun instance — Did you test it?

We should consider add it in another pr with more discuss

PR #595, opened 2019-12-31
Add basic test for a few files

Jack-Works:


scripts/jest-global-setup.js
```diff
@@ -0,0 +1,32 @@
...
+                file.close(cb)
+            })
+        })
+    }
+
+    const http = require('https')
+    const fs = require('fs')
+
+    const dist = require('path').join(__dirname, '../npm-debug.log.asmcrypto.js')
+    if (fs.existsSync(dist) && fs.readFileSync(dist, 'utf-8').length > 200) return cb()
```
SunriseFox: Maybe we can use `statSync`?...

PR #598, opened 2019-12-31
chore: provide consistent extension id when install from source

### LINKED ISSUE #28: Provide a stable extension id for users that install from source and for development

SunriseFox:
The ID will be `jkoeaghipilijlahjplgbfiocjhldnap`. Same as the one of chrome store.

? what about firefox?

close #28


neruthes:
Awesome!

PR #741, opened 2020-02-06
chore: switch to use-subscription

Jack-Works:
https://github.com/facebook/react/tree/master/packages/use-subscription

Better than our self implemented `useValueRef`


septs:
simplify code style
```typescript
const subscription = useMemo<Subscription<T>>(
    () => ({
        getCurrentValue: () => ref.value,
        subscribe: ref.addListener.bind(ref),
    }),
    [ref],
)
```

PR #796, opened 2020-02-24
fix: syntax errors in i18n msg

### LINKED ISSUE #790: [Bug] Use 'cannot' instead of 'can not'
# Bug Report

https://github.com/DimensionDev/Maskbook/blob/66161f09feec2160ef270484fca53776b9b03bd6/src/_locales/en/messages.json#L198

### System

- OS:
- OS Version:

### Browser

- Browser:
- Browser Version:

### Maskbook

- Maskbook Version:

### Expected Behavior

### Actual Behavior

### How To Reproduce

/* Specify how it may be produced here. */

Jack-Works:
close #790


Jack-Works:
@yisiliu

src/_locales/en/messages.json
```diff
@@ -195,11 +195,11 @@
     "set_up": "Set up",
     "settings": "Settings",
     "settings_disable_new_background_tab": "Disable open hidden tabs in the background",
-    "settings_disable_new_background_tab_desc": "This is an deprecated behavior of Maskbook. If you find that you can not decrypt some old posts, deselect this.",
+    "settings_disable_new_background_tab_desc": "This is a deprecated behavior of Maskbook. If you find that you cannot decrypt some old posts, deselect this.",
```
neruthes: I advised that all wordings should be in the positive tone unless it is a button for an action like "unsubscribe". If there is a switch for email notification, it should be "Receive email notifications" instead of "Disable email notifications". Imagine when you enable an option which disables something. Therefore, it will be better to write "Deprecated behavior. Enable if you need to decrypt ancient posts (created before Jan 2020)." and make its default value `false`, where the user can enable this option to enable the behavior.

src/_locales/en/messages.json
```diff
@@ -216,27 +216,27 @@
...
-    "welcome_1a1_didntfind": "Didn't find your account?",
-    "welcome_1a1_title": "Setup which account?",
+    "welcome_1a1_didntfind": "Didn't you find your account?",
+    "welcome_1a1_title": "To set up which account?",
     "welcome_1a1b_connect": "Connect",
     "welcome_1a1b_danger": "Censored network",
     "welcome_1a1b_switch": "Use existing accounts",
     "welcome_1a1b_title": "Maskbook supports these social networks",
-    "welcome_1a2_description": "Then only people you select with Maskbook can see the posted content.",
+    "welcome_1a2_description": "Then, only friends you select with Maskbook can see the posted content.",
```
neruthes: It is a good time to start reducing the mentions of "select". Manual recipient selection is optional and not mandatory. I suggest "Then, only your friends can see the content via Maskbook. (You have the option to select specific friends, too.)" for this field.

src/_locales/en/messages.json
```diff
@@ -216,27 +216,27 @@
...
     "welcome_1a3_description1": "Our extension is asking you to download your key as a file.",
     "welcome_1a3_description2": "Back it up often.",
     "welcome_1a3_done_button": "Got it!",
     "welcome_1a3_title": "Keep your backups carefully",
-    "welcome_1a3a_connect_subtitle": "will use same crypto key",
+    "welcome_1a3a_connect_subtitle": "will use the same CryptoKey",
     "welcome_1a3a_connect_title": "Connect other accounts",
     "welcome_1a3a_generate_new_key": "Generate",
-    "welcome_1a3a_generate_new_key_password_help": "Please remember the key, or your key will not be able to restore again.",
+    "welcome_1a3a_generate_new_key_password_help": "Please remember the key, or your key will not be able to restore.",
```
neruthes: I suggest "Please remember the key carefully. There is no other way to restore it.".

src/_locales/en/messages.json
```diff
@@ -257,7 +257,7 @@
     "welcome_1a4_type_auto_subtitle2": "This allows your friends to verify the connection between your social network account and your keypair.",
     "welcome_1a4_type_auto_switch": "Prefer doing it manually?",
     "welcome_1a4_type_manual_goto": "Copy & Go to Profile",
-    "welcome_1a4_type_manual_subtitle1": "Add this to bio, or post on timeline, before your first encrypted post.",
+    "welcome_1a4_type_manual_subtitle1": "Add this to your bio, or post on your timeline, before your first encrypted post.",
```
neruthes: This is related to proof post? If so, now it should be "Add this to your bio before your first encrypted post.".

src/_locales/en/messages.json
```diff
@@ -266,7 +266,7 @@
     "welcome_1b_hint_identity": "This backup file contains the following accounts:",
     "welcome_1b_hint_network": "And Maskbook is going to request permissions for these sites:",
     "welcome_1b_no_file_selected": "Select your exported key backup",
-    "welcome_1b_qr_error_1": "There is an error occur during the scanning.",
+    "welcome_1b_qr_error_1": "There is an error that occurs during scanning.",
```
neruthes: I suggest "An error occurred when scanning.".

src/_locales/en/messages.json
Tedko: 'keep'/'store' -> 'remember' ? @yisiliu

yisiliu:
Please don't merge this until I approve it. I am still reviewing it.

src/_locales/en/messages.json
```diff
@@ -1,95 +1,96 @@
...
-    "dashboard_import_backup": "Import Data Backup",
-    "dashboard_import_database_hint": "From a previous database backup.",
+    "dashboard_import_backup": "Import a data backup",
+    "dashboard_import_database_hint": "Restore from a previous database backup.",
     "dashboard_import_persona_failed": "Your import data is invalid. Please check again.",
-    "dashboard_import_persona_hint": "From a previous persona backup.",
+    "dashboard_import_persona_hint": "Restore from a previous persona backup.",
     "dashboard_imported_persona": "Imported persona `{{name}}` with {{count}} profiles.",
-    "dashboard_init_header": "Post on social networks without allowing the corporations to stalk you.",
+    "dashboard_init_header": "Post on social networks without allowing the platforms to stalk you.",
```
neruthes: One minor note: We are technically using "platform" and "network" to refer to Chrome etc and Facebook etc. We should get a consensus on terminology, but I will leave this message as it is right now.

PR #808, opened 2020-02-26
Update issue templates

neruthes:
Now using ISO week day for date notation per the advice by @septs.

### DIFF
--- a/.github/ISSUE_TEMPLATE/demand.md
+++ b/.github/ISSUE_TEMPLATE/demand.md
@@ -1,6 +1,6 @@
 ---
 name: Demand
-about: "PRD for new demand"
+about: PRD for new demand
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/demand.md
```diff
@@ -13,9 +13,9 @@ Field | Value
 --- | ---
 Previous Stage | ...
 Figma Link | [figma.com/...](___)
```
Jack-Works: In this line, use `/** figma.com/... */ Link here` might be better.

neruthes: Good point. I will change it to a simple `__Figma_Link_Here__`.

PR #833, opened 2020-03-03
fix: cannot do automation on iOS

Jack-Works:
This PR need native code from iOS @MainasuK 

Introduce a new field in the `manifest.json`: `iOS-injected-scripts` (string[]).

Maskbook iOS wrapper should inject the specified files by UserScript.

This is a Maskbook specified change, no changes should be applied on the WebExtension shim


MainasuK:
Only one constraint: no duplicate filename. For example, this is not allowed.
```
js/abc.js
script/abc.js
```

Jack-Works:
@SunriseFox does it (automatic paste) work now?

SunriseFox:
Work

PR #860, opened 2020-03-06
perf: adopt prebuilt for ios

Jack-Works:
please update to latest webextension-shim to test this pr


Jack-Works:
does it fast?🤔 @SunriseFox

SunriseFox:
It is fast but it doesn't work...

Tedko:
h😂

Jack-Works:
@SunriseFox 👀 this pr has been rebased from ios-upgrade and picked commits from #881 I think we can test the performance and functionality on iOS now...?

Jack-Works:
redir to #885

PR #914, opened 2020-03-26
fix: remove permission request on iOS

Jack-Works:


Tedko:
Let’s do an iOS release along with the wallet PR one?

PR #916, opened 2020-03-27
fix: workaround test in China

septs:


Jack-Works:
No that's weird. Developer should have their proxy setup.

Jack-Works:
And this file only need to be download once. If you don't have it, you can manually download it and store it in the right place.

PR #962, opened 2020-04-10
if empty wallets then create a default wallet else noop

septs:
https://github.com/DimensionDev/Maskbook-Talks/issues/30


Tedko:
Good idea. Lets do it

PR #969, opened 2020-04-11
chore: simplify qr-code-scan

septs:


src/utils/hooks/useQRCodeScan.tsx
```diff
@@ -1,20 +1,15 @@
 /// <reference path="../../components/QRScanner/ShapeDetectionSpec.d.ts" />
 /** This file is published under MIT License */
+import _ from 'lodash'
```
Jack-Works, line 3: Please use lodash-es and tree shake friendly import

src/utils/hooks/useQRCodeScan.tsx
```diff
@@ -1,20 +1,15 @@
...
-    const devices = (await navigator.mediaDevices.enumerateDevices()).filter((devices) => devices.kind === 'videoinput')
-    const back = devices.find(
-        (device) =>
-            (device.label.toLowerCase().search('back') !== -1 || device.label.toLowerCase().search('rear') !== -1) &&
-            device.label.toLowerCase().search('front') === -1,
-    )
-    if (back) return back.deviceId
-    if (devices[0]) return devices[0].deviceId
-    return null
+    const devices = filter(await navigator.mediaDevices.enumerateDevices(), ['kind', 'videoinput'])
```
Jack-Works, line 10: Is the 'kind' here type safe? In other words, if it get misspelled to 'kinds' or other non-existent property, does the compiler give the type error?

septs: The is lodash iteratee shorthand.

No Type safe.

PR #1011, opened 2020-04-27
Disable auto share when importing backup

Jack-Works:


neruthes:
Can someone offer some background on this decision?

Jack-Works:
> Can someone offer some background on this decision?

@Tedko reports when he is setting up a new Maskbook with his backup file, the Maskbook behaves very slow cause it is doing the heavy crypto calculation. I think auto-share during restoring the database is unwanted behavior.

neruthes:
The real solution should be a better scheduling mechanism to soften the calculation burden.

At this moment I will not block this PR. We shall design the automatic sharing mechanism ASAP.

Jack-Works:
When importing a new database, the data in the database is already shared to it's targets so it's no need to autoshare it again. @neruthes

Tedko:
LGTM.

PR #1012, opened 2020-04-27
fix: ban pinned property cause ff doesnt support it

Jack-Works:


Jack-Works:
i'll reopen a new one

PR #1086, opened 2020-05-13
New decryption UI

Jack-Works:
image


src/social-network-provider/twitter.com/ui/injectPostInspector.tsx
```diff
@@ -1,17 +1,42 @@
...
+                    parent.style.height = '0'
+                    parent.style.overflow = 'hidden'
+                }
             }
         },
     })(current)
 }
+function matches(str: string) {
+    return str.includes('maskbook.com') && str.includes('Make Privacy Protected Again')
+}
```
septs, line 35:
```suggestion
function matches(input: string) {
    return /maskbook\.com/i.test(input) && /Make Privacy Protected Again/i.test(input)
}
```

use regex pattern ignore case sensitive

PR #1099, opened 2020-05-16
Supports confirmation event in web3 api

clearloop:
## Spec

The `receipt` web3 event we are currently using in the `wallet` plugin has `reverted` risk in some situations, we can use `confirmation` instead for the more exactly UX.

## Tests

For the tests, tried upgrading jest to `v26.0.0`, but not works.


src/plugins/Wallet/real.ts
```diff
@@ -47,8 +48,37 @@ async function sendTx<R, T extends TransactionObject<R>>(txObject: T, tx: Tx = {
...
+                /**
+                 * @returns {Number, TxReceipt}
+                 *
+                 * Is fired for every confirmation up to the 12th confirmation.
+                 * Receives the confirmation number as the first and the receipt
+                 * as the second argument.
+                 *
+                 * Fired from confirmation 0 on, which is the block where its minded.
+                 */
+                .on('confirmation', (no: number, receipt: TxReceipt) => listeners?.onConfirmation?.(no, receipt))
```
guanbinrui, line 73: This event will emit multi-times, which may cause `onConfirmation` to be invoked many times and encounter the error of resolving a resolved promise.

clearloop, line 73: From the [White Paper][wp]:

> Ethereum implements a simplified version of GHOST which only goes down seven levels. Specifically, it is defined as follows:
> 
> + A block must specify a parent, and it must specify 0 or more uncles
> + An uncle included in block B must have the following properties:
> + It must be a direct child of the k-th generation ancestor of B, where 2 <= k <= 7.
> + ...

Dual to the possibility of uncle block, 7 confirmations are safety indeed, the `confirmation` event emits at the first `12th`(every 12th), it is closed to be confirmed.

Another ref from the [stackexchange][se].

---

The other way is adding a listener keeping requesting Ethereum till the block is totally mined, but it seems costs a lot, and the logic of wallet will be more complex.

[wp]: https://github.com/ethereum/wiki/wiki/White-Paper#modified-ghost-implementation
[se]: https://ethereum.stackexchange.com/questions/319/what-number-of-confirmations-is-considered-secure-in-ethereum

yisiliu:
It there a better way to get the latest gas price, other than https://ethgasstation.info/? It is good but it has a rate limit of 1000 queries each month, equivalent to 33 queries a day. I am not sure if updating every 43 minutes is steady enough to ensure each tx won't result in a gas price underpay issue.

FYI I am talking about the microservice I built to pay for the users without ETH in their account.

clearloop:
> It there a better way to get the latest gas price

+ https://etherscan.io/gastracker 
+ https://etherscan.io/apis#gastracker

Or a microservice keeps analyzing the `gastracker`?

> FYI I am talking about the microservice I built to pay for the users without ETH in their account.

Reasonable.

src/plugins/Wallet/real.ts
```diff
@@ -47,8 +48,37 @@ async function sendTx<R, T extends TransactionObject<R>>(txObject: T, tx: Tx = {
                     gas,
                     gasPrice,
                 })
+                /**
+                 * @returns {String}
```
Jack-Works, line 52: generally you don't need a JSDoc type annotation cause the type system do the work

Tedko:
LGTM

Tedko:
@clearloop need to fix the conflicts though

clearloop:
> @clearloop need to fix the conflicts though

Should I merge master into this branch? Let me try

Tedko:
o👌

PR #1102, opened 2020-05-18
feat: add gun server

septs:


src/network/gun-servers.ts
```diff
@@ -1,2 +1,2 @@
-export const gunServers = ['https://gun.maskbook.com/gun'] as const
+export const gunServers = ['https://hk.gun.dimension.chat/gun', 'https://gun.maskbook.com/gun'] as const
```
Jack-Works, line 1: This will slow down the gun. Gun will ensure the data do sync to both two server instead of randomly choose one and let the server sync by themself

septs: by design?
https://github.com/amark/gun/issues/309

septs: image

septs: random selected a server?

Jack-Works: have no idea

PR #1089, opened 2020-05-14
fix: build without git, close #1066

Jack-Works:

PR #1127, opened 2020-05-22
fix: reproducible build (#1066, #1089)

Jack-Works:


webpack.config.js
```diff
@@ -73,7 +73,18 @@ module.exports = (argvEnv, argv) => {
     const dist = env === 'production' ? src('./build') : src('./dist')
 
     const gitInfo = () => {
-        if (target.ReproducibleBuild) return undefined
```
septs, line 76: plz use `@nice-labs/git-rev@3.4.2` new api
```typescript
isRepository(): boolean
```

septs:
```typescript
const environment = { ... }
if (!target.ReproducibleBuild && git.isRepository()) {
    Object.assign(environment, { ... })
}
return new webpack.EnvironmentPlugin(environment)
```

src/plugins/Wallet/real.ts
```diff
@@ -460,6 +487,14 @@ export const walletAPI = {
                         txHash = hash
                     },
                     onReceipt() {
+                        if (receipt) {
```
septs:
```suggestion
                        if (!receipt) return
```

Jack-Works: @clearloop

clearloop: y

PR #1159, opened 2020-05-30
feat: switch to github packages

septs:


septs:
Thank you for being here! Currently you need to authenticate to download both public and private packages I’m afraid. We are investigating how, going forwards, beyond the beta we maybe able to offer this functionality, however this could be a way off yet - I’ve added a +1 against this for you, and we’ll let you know as soon as we have an update that we can share surrounding this. Keep an eye on https://github.blog 6 for all updates meanwhile.

https://github.community/t/download-from-github-package-registry-without-authentication/1440

PR #1245, opened 2020-06-16
improvement: added dependabot config to update the frequency to monthly

yisiliu:


yisiliu:
validated at https://dependabot.com/docs/config-file/validator/

yisiliu:
Merged.

PR #1302, opened 2020-07-06
fix: chrome hint `$schema` invaild field

septs:


.vscode/settings.json
```diff
@@ -1,12 +1,21 @@
 {
     "typescript.tsdk": "node_modules/typescript/lib",
     "editor.formatOnSave": true,
+    "editor.codeActionsOnSave": {
+        "source.fixAll": true
```
Jack-Works, line 5: Code action is not guaranteed to be safe. If you auto safe it, it might include wrong fixes.

septs: I think it is think too much.

Jack-Works: IIRC React dependencies array will give code fixes but it might be wrong.

PR #1338, opened 2020-07-21
Encrypting Images

nazariyv:
Would be very grateful if you could help me resolve that issue I talked about in our dimension chat

Thank you


src/social-network-provider/facebook.com/tasks/uploadToPostBox.ts
```diff
@@ -123,34 +132,40 @@ const shuffle = ({ file, blockWidth }: shuffleArgs) => {
     }
 
     const totalBlocksNum = (w * h) / (blockWidth * blockWidth) // this will be a whole number, because we resize earlier
-    const permutations: number[] = []
+    // const permutations: number[] = []
     const img = file.bitmap.data
+    const seed = String(Math.random()) // ! how many states can this generate?
```
nazariyv, line 137: a lot. 0 - 9 for each decimal place digit & 16 decimal places => 10 ^ 16 states of the prng. That means, that log base 2 of 10 ^ 16 = 53.15. So the prng can generate 2 ^ 53.15 states. Solid

Tedko:
@yisiliu @Jack-Works @guanbinrui

src/components/InjectedComponents/PostDialog.tsx
```diff
@@ -16,12 +16,15 @@ import {
     DialogProps,
     Tooltip,
 } from '@material-ui/core'
+import Jimp from 'jimp'
```
septs, line 19: why not use `ImageData`.

https://mdn.io/ImageData

It is better if you can reduce the extra dependency

septs:
It is a better choice to extract this encryption method into a separate library, thanks

like https://github.com/DimensionDev/Stego-JS, https://github.com/DimensionDev/common-protocols
and publish to github packages (https://github.com/orgs/DimensionDev/packages)

nazariyv:
of course, I will do that in the end. First, I would like to know why the shuffled image isn't being uploaded to the post box in Facebook

src/components/InjectedComponents/PostDialog.tsx
nazariyv: of course, will change

nazariyv: I wasn't aware of it

nazariyv:
actually, I won't be able to do that. I am properly running low on time now, and need to get this finished ASAP and record a video about it. Which will take another day, I need to edit it too.

nazariyv:
I am very sorry if I come across rude, I am not and I do not intend to be. Just please understand that I am planning things out so that I can efficiently spend time on everything that I do, and this is taking a bit too long now

nazariyv:
Thinking behind this code. For the lack of a better place to share it. I will keep it here for reference.

Some of the things written there have issues, but the code doesn't. Apart from the off by one pixel block

Hackathons-2
Hackathons-3
Hackathons-4
Hackathons-5
Hackathons-6
Hackathons-7

Tedko:
> I am very sorry if I come across rude, I am not and I do not intend to be. Just please understand that I am planning things out so that I can efficiently spend time on everything that I do, and this is taking a bit too long now

@nazariyv No worry. This is a tricky bounty and also our first bounty. I guess our engineer will help you out in around 1 day and we can count this bounty as achieved once we uploaded the 1st encrypted img and decrypt it successful! 
Also we will init a new repo for this lib under https://github.com/DimensionDev/ and ask you to pr there then :) Welcome to join community devs of Maskbook :) 🥇

nazariyv:
All of that works, I have tested it on my FB, but without integrating it into Maskbook :) Where it doesn't work is where I paste into the active element, that custom event part doesn't upload the image 😔

guanbinrui:
I'm going to integrate it into Maskbook today.

nazariyv:
I have just fixed the issue

nazariyv:
you have a race condition somewhere. but if you are here. you can help me with the final bit - I need to pull out the image in the postPayload

nazariyv:
just pushed the bit that fixes the race condition

nazariyv:
once I have the image, I will call deshuffle and then need to inject in place of the shuffled image

src/components/InjectedComponents/PostDialog.tsx
```diff
@@ -16,12 +16,15 @@ import {
...
+import type JimpT from 'jimp'
+var toBuffer = require('typedarray-to-buffer') // TODO: !
```
Jack-Works, line 21: Hmm todo what?

src/components/InjectedComponents/PostDialog.tsx
```diff
@@ -89,21 +98,41 @@ export interface PostDialogUIProps
...
+    onImageEncryptSwitchChanged: (checked: boolean) => void
     onFinishButtonClicked: () => void
     onCloseButtonClicked: () => void
     onSetSelected: SelectRecipientsUIProps['onSetSelected']
     DialogProps?: Partial<DialogProps>
     SelectRecipientsUIProps?: Partial<SelectRecipientsUIProps>
 }
+
+// TODO: you will want to move this elsewhere
+const readFileAsync: (file: File) => Promise<ArrayBuffer | null> = (file: File) => {
```
Jack-Works, line 121: Looks familiar, we have the same logic at somewhere else. we can make it into utils

src/components/InjectedComponents/PostDialog.tsx
```diff
@@ -89,21 +98,41 @@ export interface PostDialogUIProps
...
+        const reader = new FileReader()
+        reader.onload = () => {
+            if (typeof reader.result === 'string') {
+                resolve(null)
+            } else {
+                resolve(reader.result)
+            }
+        }
+        reader.onerror = reject
+        reader.readAsArrayBuffer(file)
```
septs, line 132: why not `await blob.arrayBuffer()`?

see https://mdn.io/blob/arrayBuffer

src/components/InjectedComponents/PostDialog.tsx
septs:
```
Buffer.from(uint8array)
```

it work

nazariyv: I have never used that Node.js style import in a React project before. But I have read that this gets converted into the right syntax with babel. Right?

septs: webpack bundler include `Buffer` polyfill.
see https://v4.webpack.js.org/configuration/node/

septs: but webpack 5 removed the support
see https://github.com/webpack/changelog-v5#changes-to-the-structure

septs: but, note `.slice(...)`  is reference parent buffer

```plain
> payload = Buffer.from('example-1', 'ascii')
<Buffer 65 78 61 6d 70 6c 65 2d 31>
> sliced = payload.slice(8)
<Buffer 31>
> sliced.write('2')
1
> payload.toString()
'example-2'
>
```

guanbinrui:
Hi @nazariy, I did some refactoring on your code. Plz check the `feature/encrypt-image` branch. Your shuffle code was moved to the `ImageShuffleService`. For now, it de-shuffles all images and presents each result for debugging purposes. I have broken your code when I was refactoring it. It always returns a noisy image. Can u give me a hand to fix it?

some updates u may want to know:

1. the pkg `jimp` and `typedarray-to-buffer` were removed, so I’m going to find a replacement way for resizing the image.
2.  u have to do testing with an image whose width is dividable by the block size (caused by 1)
3. a fixed seed was used

a preview:

preview

septs:
> 1. for resizing the image.

why not use **Canvas**?

in node, use https://npm.im/canvas
in browser, use Canvas OR OffscreenCanvas

nazariyv:
@guanbinrui I will have a look today. thank you!

src/components/InjectedComponents/PostDialog.tsx
```diff
@@ -349,6 +399,26 @@ export function PostDialog(props: PostDialogProps) {
                         template: isRedPacket ? (isDai ? 'dai' : isOkb ? 'okb' : 'eth') : 'v2',
                         warningText: t('additional_post_box__steganography_post_failed'),
                     })
+                } else if (imageEncryptEnabled) {
+                    if (!imgToEncrypt) return
+
+                    // TODO:
+                    // use dynamic seed
+                    const seed = '7380309746363496'
+                    const seedTypedMessage = makeTypedMessage('text', seed)
```
Jack-Works, line 408: strange. are you creating a non-sense text message?

src/components/InjectedComponents/TypedMessageRenderer.tsx
```diff
@@ -62,6 +70,19 @@ export const DefaultTypedMessageTextRenderer = React.memo(function DefaultTypedM
...
 
+export const DefaultTypedMessageImageRenderer = React.memo(function DefaultTypedMessageImageRenderer(
+    props: TypedMessageRendererProps<TypedMessageImage>,
+) {
+    console.log(props.message.content)
+    return renderWithMetadata(
+        props,
+        <Typography variant="body1" style={{ lineBreak: 'anywhere' }} data-testid="text_payload">
+            {/* <RenderText text={props.message.content}></RenderText> */}
+            <img style={{ width: '100%' }} src={`data:image/png;base64,${props.message.content}`} />
```
Jack-Works, line 81: Notice: data url might be blocked to render. please use a canvas to draw the image

src/extension/background-script/CryptoServices/utils.ts
```diff
@@ -28,22 +28,41 @@ export interface TypedMessageText extends TypedMessageMetadata {
...
-export function makeTypedMessage(content: string, meta?: ReadonlyMap<string, unknown>): TypedMessage {
-    if (typeof content === 'string') {
-        const text: TypedMessageText = { type: 'text', content, version: 1, meta }
-        return text
-    }
-    const msg: TypedMessageUnknown = { type: 'unknown', version: 1, meta }
-    return msg
+export type TypedMessage = TypedMessageText | TypedMessageImage | TypedMessageComplex | TypedMessageUnknown
+export function makeTypedMessage(
+    type: 'text' | 'image' | 'complex' | 'unknown',
```
Jack-Works, line 44: I prefer a factory function for each type cause their construction argument may variant a lot.

neruthes:
I have been aware of that there is a "seed placement" question here. I wonder if it is possible to put the random seed as the first line of 16x16 chunks in the image; I wish that the image can be decoupled from the text. 

For example:

- The initial 2 chunks are pure Cyan and pure Magenta. This is for version control and pattern match.
- The next chunks can encode the actual data. Each chunk contains 4 subchunks (4x4). Each subchunk can be 000, F00, 0F0, or 00F.

This example would make the clues for permutations more visible, but seemingly does not introduce extra visibility for machines?

nazariyv:
I am struggling to meaningfully answer this question, since it is not clear what Facebook does to the image. Although, we do have two clues

1. when we shuffle by pixel (4 bytes), then after we download the img from Facebook and deshuffle, we get greyscale version
2. when we use at least 2 x 2 blocks, colours are preserved

It is important to know what happens to the image because it can ruin the seed embedded into the image itself. And if we were to go this route, I would also salt the seed with something, and then store that in a chunk instead.

What is the problem with having an encrypted seed? In that the user may type up some other text? I suppose then this also relates to the key that we are saving for users in the bio, too. We can solve this problem more simplistically right now, by having the same mechanism for this seed (i.e. just separate it out with emojis from the rest of the text)

guanbinrui:
replaced by #1352

src/components/InjectedComponents/TypedMessageRenderer.tsx
guanbinrui: Should be done by another PR.

src/components/InjectedComponents/PostDialog.tsx
guanbinrui: Not supported by (iOS) safari.

neruthes:
> What is the problem with having an encrypted seed? In that the user may type up some other text?

Things are tricky here. In the original design of Plugins, a Plugin may not read or write a Natural Post (SNS-defined post), but should only add Plugin-Specific Metadata to the "Masked Post" (including pre-decryption metadata and post-decryption metadata).

If we are going to let a Plugin write data directly to the Natural Post ("pass-through"), as discussed in #1360, it will be difficult to find a way to let it take advantage of the recipient-control feature of Masked Post, while it can read and write arbitrary data from/into Natural Posts.

For now, since the permutation is chunk-based and not bit-based, the major purpose could be countering automated censorship algorithms instead of prepared analyzers. With this in mind, may I suppose that it should be fine to write the unencrypted seed into the Natural Post?

src/components/InjectedComponents/PostDialog.tsx
Jack-Works: Let's add a polyfill for this. #1355 is using Blob.arraybuffer too.

PR #1347, opened 2020-07-27
feat: add api-extractor

Jack-Works:
await microsoft/rushstack#1796


Jack-Works:
We're still blocking by [microsoft/rushstack#1796](https://github.com/microsoft/rushstack/pull/1796) but I'm closing this PR because I need to implement this in the base of the master branch instead of the ESM branch.

PR #1355, opened 2020-07-29
canvas drawer

Jack-Works:


.storybook/main.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
         '@storybook/addon-viewport/register',
     ],
     webpackFinal: async (/** @type {import('webpack').Configuration} */ config) => {
+        config.output
```
guanbinrui, line 13:
```suggestion
```

src/components/shared/Image.tsx
```diff
@@ -0,0 +1,159 @@
...
+     */
+    component?: 'img' | 'canvas'
+    width: number
+    height: number
+    loading?: boolean
+    // usability
+    canvasProps?: React.DetailedHTMLProps<React.CanvasHTMLAttributes<HTMLCanvasElement>, HTMLCanvasElement>
+    imgProps?: React.DetailedHTMLProps<React.ImgHTMLAttributes<HTMLImageElement>, HTMLImageElement>
+    SkeletonProps?: SkeletonProps
+    className?: string
```
guanbinrui, line 36: Remove the `className` property. The user can do the same thing with `canvasProps` or `imgProps`. BTW, better use `CanvasProps` and `ImgProps`.

src/components/shared/Image.tsx
```diff
@@ -0,0 +1,159 @@
...
+        async function () {
+            if (propsLoading) return
+            if (component === 'img') return
+            if (typeof src !== 'string') return
+            if (origin === 'current') return fetch(src).then((x) => x.blob())
+            return Services.Helper.fetch(src)
+        },
+        [component, src, origin],
+    )
+    if (error) console.error(error)
```
guanbinrui, line 72: Should not continue if fetch failed.

src/utils/utils.ts
```diff
@@ -26,7 +26,7 @@ export function getUrl(path: string, fallback: string = '') {
 }
 
 /**
- * Download given url return as ArrayBuffer
+ * Download given url return as Blob
  */
 export async function downloadUrl(url: string) {
```
guanbinrui, line 31: Add a new methods `downloadUrlAsArrayBuffer` or `downloadUrlAsBlob`?

src/components/shared/Image.tsx
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * This React Component is used to render images in the content script to bypass the CSP restriction.
+ */
+export const Image = forwardRef<ForwardingRef, ImageProps>(function Image(props, outgoingRef) {
+    const { src, loading: propsLoading, canvasProps, imgProps, style, className, SkeletonProps } = props
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/canvas#Maximum_canvas_size
+    const [height, width] = [Math.min(32767, props.height), Math.min(32767, props.width)]
+    const [origin, component] = resolveMode(props)
+    const canvasRef = useRef<HTMLCanvasElement>(null)
+    const imgRef = useRef<HTMLImageElement>(null)
```
guanbinrui, line 55:
```suggestion
    const canvasRef = useRef<HTMLCanvasElement | null>(null)
    const imgRef = useRef<HTMLImageElement | null>(null)
```

src/components/shared/Image.tsx
```diff
@@ -0,0 +1,159 @@
...
+        return await createImageBitmap(arr)
+    } catch {
+        // Safari route
+        return new Promise<CanvasImageSource>((resolve, reject) => {
+            const img = document.createElement('img')
+            img.addEventListener('load', function () {
+                resolve(this)
+                URL.revokeObjectURL(img.src)
+            })
+            img.addEventListener('error', reject)
```
guanbinrui, line 154: Haven't revoke object URL in the error branch.

src/components/shared/Image.tsx
Jack-Works: it's provided for ergonomic reason

src/components/shared/Image.tsx
Jack-Works: IDK what should be returned in this case.

src/utils/utils.ts
Jack-Works: I think it isn't necessary. It's easy to get ArrayBuffer from Blob

src/components/shared/Image.tsx
Jack-Works: No need to change. It's already type-safe.

src/extension/background-script/SteganographyService.ts
```diff
@@ -47,7 +47,10 @@ const defaultOptions = {
 const isSameDimension = (dimension: Dimension, otherDimension: Dimension) =>
     dimension.width === otherDimension.width && dimension.height === otherDimension.height
 
-const getMaskBuf = memoizePromise((type: Mask) => downloadUrl(getUrl(`/image-payload/mask-${type}.png`)), undefined)
+const getMaskBuf = memoizePromise(
+    (type: Mask) => downloadUrl(getUrl(`/image-payload/mask-${type}.png`)).then((x) => x.arrayBuffer()),
```
septs, line 51:
```suggestion
    (type: Mask) => (await downloadUrl(getUrl(`/image-payload/mask-${type}.png`))).arrayBuffer(),
```

Jack-Works: that's the same

src/components/shared/Image.tsx
guanbinrui: Maybe a default image as a placeholder when the error happens.

src/extension/background-script/SteganographyService.ts
guanbinrui: It seems weird to me that mixing `await` and `.then`.

src/components/shared/Image.tsx
guanbinrui: It's better to design like MUI components. It will be easier for someone else who uses this component. I still advise using a big camel style as the properties' name.

Jack-Works: there are no inner DOM structures so `*Props` is enough (no need to use `classes`)

src/components/shared/Image.tsx
Jack-Works: what should it be?

guanbinrui: A Maskbook logo with opacity 0.1 which is placed in a light gray background. 🤔

src/components/shared/Image.tsx
Jack-Works: > BTW, better use `CanvasProps` and `ImgProps`.

The naming is also following MUI style cause `img` and `canvas` are native elements meanwhile `Skeleton` are React component

src/components/shared/Image.tsx
Jack-Works: We'd better leave it in future

Jack-Works:
CC @septs

PR #1364, opened 2020-07-31
Plugin db (picked from #1331)

Jack-Works:
The new plugin db design:

*   give up the isomorphism to the Indexed DB (it's too complex)
*   give up the consistency guarantee by the transaction (reason: it's not possible to share a transaction in the external wallet format)
*   use a simpler & minimal API set that works like `localStorage`
*   MUST use tagged union to store the data
*   old plugin DB is deprecated. migrate in future

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,5 +17,8 @@
             "fileMatch": ["/src/manifest.json"],
             "url": "http://json.schemastore.org/chrome-manifest"
         }
+    ],
+    "cSpell.words": [
+        "Reificated"
     ]
 }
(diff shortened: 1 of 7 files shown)


src/plugins/FileService/database.ts
```diff
@@ -0,0 +1,4 @@
+import { createPluginDatabase } from '../../database/Plugin/wrap-plugin-database'
+import { identifier } from './constants'
+
+export const FileServiceDatabase = createPluginDatabase(identifier)
```
septs, line 4: remove the file

PR #1396, opened 2020-08-03
Fix Japanese Locale

blank71:


neruthes:
Thanks for the pull request. I believe that this will be better than the existing translation and am wondering if "テーマ" could be even better. Any idea on this?

blank71:
Great. I think "テーマ" is more familiar than "外観".

Tedko:
Ok. Do you mind do another PR on テーマ ?

blank71:
https://github.com/DimensionDev/Maskbook/pull/1397

PR #1397, opened 2020-08-03
Fix Japanese Locale

blank71:
"外観" to "テーマ"

### DIFF
--- a/src/_locales/ja/messages.json
+++ b/src/_locales/ja/messages.json
@@ -312,7 +312,7 @@
     "settings_enable_debug_desc": "デバッグのための追加情報を参照してください。",
     "settings_image_based_payload": "イメージモードを有効にする",
     "settings_image_based_payload_desc": "テキストの代わりに画像のペイロードで投稿を公開します。",
-    "settings_appearance": "外観",
+    "settings_appearance": "テーマ",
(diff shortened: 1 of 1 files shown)


Tedko:
@blank71 A quick question:

Is this a good translation of `Image Payload`?


Screen Shot 2020-08-08 at 17 11 51


Note: This is not encrypted a picture into something else. But `hide` encrypted texts/red packets into an Image.

blank71:
@Tedko I'm not a software engineer or something like that so I'm not familiar to the word 'Image Payload' but I feel this translation is good.
But '画像に暗号化' is more better for me.

’画像で暗号化’ also means that use image to output encrypted thing. If '画像に暗号化' that means output is a image. So the latter is better.😊

By the way, if you intend encrypt a picture into something else it should be '画像を暗号化'.

Tedko:
> @Tedko I'm not a software engineer or something like that so I'm not familiar to the word 'Image Payload' but I feel this translation is good.
> But '画像に暗号化' is more better for me.
> 
> ’画像で暗号化’ also means that use image to output encrypted thing. If '画像に暗号化' that means output is a image. So the latter is better.😊
> 
> By the way, if you intend encrypt a picture into something else it should be '画像を暗号化'.

@blank71 great thanks!!

PR #1454, opened 2020-08-20
feature: parse original post as TypedMessage

Jack-Works:

### DIFF
--- a/src/protocols/typed-message/debugger.tsx
+++ b/src/protocols/typed-message/debugger.tsx
@@ -1,0 +1,87 @@
+// @ts-ignore
+const React = require('jsx-jsonml-devtools-renderer') as any
+import type { TypedMessage, TypedMessageCompound, TypedMessageImage, TypedMessageText } from './types'
+
+class TypedMessageFormatter {
+    isTypedMessage(obj: unknown): obj is TypedMessage {
+        if (typeof obj !== 'object' || obj === null) return false
+        if (!('version' in obj)) return false
+        if (!('meta' in obj)) return false
+        if (!('type' in obj)) return false
+        return true
+    }
+    hasBody(obj: unknown) {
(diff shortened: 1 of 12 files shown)


septs (approved):

src/protocols/typed-message/debugger.tsx
```diff
@@ -0,0 +1,87 @@
...
+        if (!this.isTypedMessage(obj)) return false
+        if (obj.type === 'empty') return false
+        return true
+    }
+    compound(obj: TypedMessageCompound) {
+        return (
+            <div style={{ maxWidth: '95vw', overflow: 'break-word' }}>
+                <ol>
+                    {obj.items.map((x) => (
+                        <li>{display(x)}</li>
```
guanbinrui, line 23: `key` is lost.

src/protocols/typed-message/debugger.tsx
```diff
@@ -0,0 +1,87 @@
...
+        return (
+            <table>
+                <tr style={{ background: 'rgba(255, 255, 255, 0.6)' }}>
+                    <td style={{ minWidth: '4em' }}>Field</td>
+                    <td>Value</td>
+                </tr>
+                {Object.keys(obj)
+                    .filter((x) => x !== 'type' && x !== 'meta' && x !== 'version')
+                    .map((x) => (
+                        <tr>
```
guanbinrui, line 39: `key`is lost.

src/protocols/typed-message/debugger.tsx
Jack-Works: This file is not React, it's a custom JSX framework. the render target does not support "key".

This file renders TypedMessage in the Chrome dev tools.

guanbinrui (approved):

PR #1489, opened 2020-08-28
chore: remove unused locale keys by scanner script

zhouhanseng:
Using current scanner to remove almost all unused locale keys
Please try [scanner script](https://github.com/zhouhanseng/locale-scaner) to verify.


septs:
@zhouhanseng plz keep original sort.

PR #1493, opened 2020-08-31
fix: locale-kit edge case

septs:


scripts/locale-kit.ts
```diff
@@ -38,13 +38,21 @@ function getUsedKeys(content: string) {
...
-                if (localeKey && (ts.isStringLiteral(localeKey) || ts.isNoSubstitutionTemplateLiteral(localeKey))) {
+                const localeKey = closest(node, ts.isCallExpression)?.arguments[0]
+                if (localeKey === undefined) {
+                    return node
+                } else if (ts.isStringLiteral(localeKey) || ts.isNoSubstitutionTemplateLiteral(localeKey)) {
                     keys.add(localeKey.text)
+                } else if (
+                    ts.isConditionalExpression(localeKey) &&
+                    ts.isStringLiteral(localeKey.whenTrue) &&
+                    ts.isStringLiteral(localeKey.whenFalse)
```
Jack-Works, line 52: Plz use `isStringLiteralLike` it can detect both string and template string

Jack-Works (changes requested):

Jack-Works (approved):

PR #1509, opened 2020-09-02
chore: improve locale kit

zhouhanseng:
Support
 
```tsx
const message = 'message'
const message1 = 'message1'
const message2 = 'message2'

t(message)

t(isTrue ? message1 : message2)

<UIComponent i18nKey={message}>
```

### DIFF
--- a/scripts/locale-kit.ts
+++ b/scripts/locale-kit.ts
@@ -41,22 +41,43 @@ function getUsedKeys(content: string) {
         }
         return undefined
     }
-    const addKey = (node: ts.Node) => {
-        if (ts.isStringLiteralLike(node)) {
-            keys.add(node.text)
-        }
-    }
     const transformer = (context: ts.TransformationContext) => (rootNode: ts.Node) => {
+        const setFromVariableWrapper = (variableValue: string): ((node: ts.Node) => ts.Node) => {
+            const setFromVariable = (node: ts.Node): ts.Node => {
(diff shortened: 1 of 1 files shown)


scripts/locale-kit.ts
```diff
@@ -41,22 +41,41 @@ function getUsedKeys(content: string) {
...
-    const addKey = (node: ts.Node) => {
-        if (ts.isStringLiteralLike(node)) {
-            keys.add(node.text)
-        }
-    }
     const transformer = (context: ts.TransformationContext) => (rootNode: ts.Node) => {
+        const setFromVariableWrapper = (variableValue: string): ((node: ts.Node) => ts.Node) => {
+            const setFromVariable = (node: ts.Node): ts.Node => {
+                if (
+                    ts.isVariableDeclaration(node) &&
```
Jack-Works, line 48: Please check if the variable declaration is a `const` declaration.

Jack-Works (dismissed):
Approved, but I think the better way is to ban the indirect usage as you mentioned in the FileService instead of trying to support those cases.

scripts/locale-kit.ts
septs:
```ts
if (ts.isVariableDeclarationList(node.parent)) {
    const constFlag = node.parent.flags & ts.NodeFlags.Const
}
```

PR #1510, opened 2020-09-02
feat: upgrade to mui-v5.0.0-alpha.5

Jack-Works:
We cannot upgrade higher than alpha 5 (iamhosseindhv/notistack#295)

We need the upgrade to at least alpha 4 cause alpha 4 improve usage with StrictMode

This PR also removes some hacks on the ShadowRoot or getComputedStyles to be more compatible with webextension shim.

One thing we need to care about is a breaking change in mui-v5.0.0-alpha4. It drops the use of React.findDOMNode so it becomes more strict on the ref forwarding and we need to check if our code is broken.

> *   \[core\] Drop support for non-ref-forwarding class components ([#21811](https://github.com/mui-org/material-ui/pull/21811)) [**@eps1lon**](https://github.com/eps1lon)  
>     Support for non-ref-forwarding class components in the `component` prop or as an immediate `children` has been dropped. If you were using `unstable_createStrictModeTheme` or didn't see any warnings related to `findDOMNode` in `React.StrictMode` then you don't need to do anything.  
>     Otherwise check out the ["Caveat with refs" section in our composition guide](https://github.com/mui-org/material-ui/blob/v5.0.0-alpha.4/guides/composition/#caveat-with-refs) to find out how to migrate.  
>     This change affects almost all components where you're using the `component` prop or passing `children` to components that require `children` to be elements (e.g. `<MenuList><CustomMenuItem /></MenuList>`)


Jack-Works:
```
    ✓ backup file - db_backup_0_persona_0_profile (4752 ms)
    ✓ backup file - db_backup_1_persona_0_profile (6754 ms)
    ✕ backup file - db_backup_1_persona_1_profile (34709 ms)
    ✕ backup file - db_backup_2_personas_0_profile (34716 ms)
    ✕ backup file - db_backup_2_personas_2_profiles (34720 ms)
    ✕ backup file - db_backup_0_persona_0_profile (34704 ms)
    ✕ backup file - db_backup_1_persona_0_profile (34705 ms)
    ✕ backup file - db_backup_1_persona_1_profile (34732 ms)
    ✕ backup file - db_backup_2_personas_0_profile (34728 ms)
    ✕ backup file - db_backup_2_personas_2_profiles (34732 ms)
```

Jack-Works:
Ok I figure out why the test failed. The ClickAwayListener seems doesn't work well...

Jack-Works:
IMO it's ready for review @guanbinrui @septs

Issue #1483, opened 2020-08-27
[Bug] Multiple payload pasted when composing encrypted post


Tedko:
# Bug Report 

As the title. easy to reproduce 
<img width="1229" alt="Screen Shot 2020-08-28 at 01 00 53" src="https://user-images.githubusercontent.com/6687437/91472602-482a7000-e8ca-11ea-9e6d-77222b9974a5.png">

### System

- [x] Windows
    - OS Version:
- [x] Mac OS X
    - OS Version:
- [x] Linux
    - Linux Distribution:
    - OS Version:

### Platform/Browser

- [x] Chrome
  - Maskbook Version: 
  - Browser Version:

- [x] Firefox
  - Maskbook Version:
  - Browser Version:

- [ ] Android
  - Maskbook Version:
  - Android Version:

### Build Variant

- Where do you get Maskbook? 
  - [ ] Store

### Actual Behavior

/* What happened? */

Jack-Works:
![image](https://user-images.githubusercontent.com/5390719/91810905-48779200-ec61-11ea-801f-34b26aac35ff.png)

Cannot reproduce on 0159f2f4

PR #1512, opened 2020-09-02
chore: remove duplicated injected script

zhouhanseng:
close https://github.com/DimensionDev/Maskbook/issues/1483
content-script would be loaded in all case, content-script contains injected-script already: https://github.com/DimensionDev/Maskbook/commit/975febc10c035ef0015089b501b428f7d842d8ea#diff-075daa7a311f82142f0742717ab50b51R1

so I think we can just remove other injected-script load.

### DIFF
--- a/src/content-script.ts
+++ b/src/content-script.ts
@@ -1,4 +1,3 @@
-import './extension/injected-script/addEventListener'
 import './setup.ui'
 import { GetContext } from '@holoflows/kit/es'
 if (GetContext() === 'content') {


Jack-Works:
> content-script contains injected-script already

Oh wait, it shouldn't, that's an error...

The correct behavior is:

- load injected script (in the extension context) only in Firefox
- load injected script in the webpage context in environments other than Firefox or Safari

Your change is not correct, the correct way is removing the code you mentioned 975febc#diff-075daa7a311f82142f0742717ab50b51R1

zhouhanseng:
Please help me squash the commits this time, `git push --force` can't work, I'm solving that problem.

septs (approved):

Issue #1488, opened 2020-08-28
[Demand] Red Packet minor text update


neruthes:
## Catalogue

- [x] UI/UX
- [ ] Maskbook Core Feature
- [ ] Maskbook Plugin Feature
- [ ] Security/Cryptography Feature
- [ ] Other
    - /* List it here */

## Part 1: Introduction and Background

![image](https://user-images.githubusercontent.com/2139261/91548560-14286c80-e915-11ea-852d-f205414afa88.png)

https://t.me/maskbook_group/10031

## Part 2: Problem to Solve

Reduce confusion.

## Part 3: Expected Behavior

Update the text template to "Expired" when the user fails to claim any share because the remaining share is 0.

## Notes

- Other ideas on the text template are welcomed.

hging:
I want to claim the task.
I will submit some changes about this page.

Tedko:
Thx @hging . Will review once u submit

PR #1521, opened 2020-09-03
feat: add some text for reduce confusion

hging:
close: #1488


Tedko (approved):

src/plugins/RedPacket/UI/RedPacket.tsx
```diff
@@ -240,6 +235,32 @@ export function RedPacket(props: { redPacket?: RedPacketRecord }) {
     )
 }
 
+function getDescription(
```
septs, line 238: @neruthes the red packet status is need i18n?

neruthes: I believe that i18n will be good

PR #1560, opened 2020-09-14
fix: cache qrcode

zhouhanseng:
fixes: #1559

### DIFF
--- a/src/components/shared/qrcode.tsx
+++ b/src/components/shared/qrcode.tsx
@@ -3,6 +3,15 @@ import { useAsync } from 'react-use'
 import qr from 'qrcode'
 import { iOSHost } from '../../utils/iOS-RPC'
 
+interface QRProps {
+    text: string
+    options?: qr.QRCodeRenderersOptions
+    canvasProps?: React.DetailedHTMLProps<
+        React.CanvasHTMLAttributes<HTMLCanvasElement & HTMLImageElement>,
+        HTMLCanvasElement & HTMLImageElement
+    >
+}
+
 const cache = new Proxy(sessionStorage, {
     get(t, p: 'get' | 'set') {
         return {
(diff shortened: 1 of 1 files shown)


src/components/shared/qrcode.tsx
```diff
@@ -16,26 +25,19 @@ const cache = new Proxy(sessionStorage, {
...
 
-export function QRCode(props: {
-    text: string
-    options?: qr.QRCodeRenderersOptions
-    canvasProps?: React.DetailedHTMLProps<
-        React.CanvasHTMLAttributes<HTMLCanvasElement & HTMLImageElement>,
-        HTMLCanvasElement & HTMLImageElement
-    >
-}) {
+export function QRCode({ text, options = {}, canvasProps }: QRProps) {
```
zhouhanseng, line 28: `qr.toCanvas()` require option shouldn't be `undefined`

src/components/shared/qrcode.tsx
```diff
@@ -16,26 +25,19 @@ const cache = new Proxy(sessionStorage, {
...
-        if (cache.get(props.text) || !ref.current) return
-        qr.toCanvas(ref.current, props.text, props.options)
-        return () => {
-            // if already rendered canvas, do not re-render img
-            // eslint-disable-next-line react-hooks/exhaustive-deps
-            cache.set(props.text, ref.current?.toDataURL())
-        }
-    }, [props.options, props.text])
-    return image ? <img src={image} {...props.canvasProps} /> : <canvas {...props.canvasProps} ref={ref} />
+        if (!ref.current) return
```
zhouhanseng, line 32: Although `useEffect` runs after every render cycle, which means we can get ref first time. But It is the case that `ref` must be `undefined` since it also returns . So `if (!ref.current) return` is still necessary .

src/components/shared/qrcode.tsx
```diff
@@ -16,26 +25,19 @@ const cache = new Proxy(sessionStorage, {
...
+
+        qr.toCanvas(ref.current, text, options, () => {
```
zhouhanseng, line 34: We should get `ref.current?.toDataURL()` in the callback function of `qr.toCanvas()` rather than useEffect's 'cleanup' return function.

src/components/shared/qrcode.tsx
```diff
@@ -16,26 +25,19 @@ const cache = new Proxy(sessionStorage, {
...
     const ref = useRef<HTMLCanvasElement | null>(null)
-    const image = cache.get(props.text)
+    const image = cache.get(text)
     useEffect(() => {
-        if (cache.get(props.text) || !ref.current) return
-        qr.toCanvas(ref.current, props.text, props.options)
-        return () => {
-            // if already rendered canvas, do not re-render img
-            // eslint-disable-next-line react-hooks/exhaustive-deps
-            cache.set(props.text, ref.current?.toDataURL())
```
zhouhanseng, line 35: Here `ref.current?.toDataURL()` is undefined, which causes this bug.

Jack-Works (approved):

src/components/shared/qrcode.tsx
```diff
@@ -16,26 +25,19 @@ const cache = new Proxy(sessionStorage, {
...
-        }
-    }, [props.options, props.text])
-    return image ? <img src={image} {...props.canvasProps} /> : <canvas {...props.canvasProps} ref={ref} />
+        if (!ref.current) return
+
+        qr.toCanvas(ref.current, text, options, () => {
+            if (ref.current?.toDataURL()) {
```
septs, line 35: plz don't do this, extract `toDataURL()` result to variable.

septs (changes requested):

src/components/shared/qrcode.tsx
zhouhanseng: Okay

src/components/shared/qrcode.tsx
```diff
@@ -16,26 +25,19 @@ const cache = new Proxy(sessionStorage, {
...
+                cache.set(text, ref.current?.toDataURL())
+            }
+        })
+    }, [options, text])
```
Jack-Works, line 39: `options` is an object which may change every time. I suggest making it JSON.stringified if it is not containing complex values

src/components/shared/qrcode.tsx
septs: storage Blob URL to cache is better choice.

big data url, will bring side effects to DOM.

you can test insert ten a 10 MiB image to DOM, try scroll bar.

src/components/shared/qrcode.tsx
zhouhanseng: Well, this `options` refers to QRcode's style and more. Currently, It is always `options={{ width: 200 }}`.

Jack-Works: If it is `options={{ width: 200 }}`, it **will** change reference **every** render

PR #1571, opened 2020-09-16
chore(new facebook): post box inject right place

zhouhanseng:
image

### DIFF
--- a/src/social-network-provider/facebook.com/UI/injectPostBox.tsx
+++ b/src/social-network-provider/facebook.com/UI/injectPostBox.tsx
@@ -13,6 +13,7 @@ if (isMobileFacebook) {
 } else {
     composeBox = new LiveSelector()
         .querySelectorAll('form [role="button"]')
+        .filter((x) => x.textContent === 'Post')
         .map((x) => x.parentElement)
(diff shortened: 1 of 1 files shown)


src/social-network-provider/facebook.com/UI/injectPostBox.tsx
```diff
@@ -13,6 +13,7 @@ if (isMobileFacebook) {
 } else {
     composeBox = new LiveSelector()
         .querySelectorAll('form [role="button"]')
+        .filter((x) => x.textContent === 'Post')
```
Jack-Works, line 16: Hmm, I'm sorry but that doesn't work. Facebook has multiple language support, if you only recognize "Post", you will miss support for other languages.

Jack-Works (changes requested):

src/social-network-provider/facebook.com/UI/injectPostBox.tsx
zhouhanseng: Update, `querySelectorAll('form [role="button"][tabindex="-1"]')` seems won't break other place and solve this.

PR #1573, opened 2020-09-16
feat: unlock uploader of file service in mobile

Jack-Works:


src/plugins/FileService/arweave/remote-signing.ts
```diff
@@ -6,7 +6,7 @@ import { signing } from '../constants'
 export async function sign(transaction: Transaction) {
     const response = await fetch(signing, {
         method: 'POST',
-        body: await makeRequest(transaction),
+        body: (await makeRequest(transaction)).buffer,
```
septs, line 9: this line need revert.

PR #1587, opened 2020-09-21
fix: setup identity ref after initializing persona (new facebook)

zhouhanseng:
Solving bug: 

image
image


src/components/InjectedComponents/ImmersiveGuide/SetupGuide.tsx
```diff
@@ -506,6 +507,7 @@ export function SetupGuide(props: SetupGuideProps) {
                 _wallet_is_default: true,
```
zhouhanseng, line 507: `Services.Identity.setupPersona()` set the `uninitialized` to `false`, so we should `MessageCenter.emit('identityUpdated', undefined)` after that finish

septs (approved):

Jack-Works (approved):

PR #1595, opened 2020-09-23
Feature: add avatar in uniswap trade view

hging:
add avatar in uniswap trade view
fix wrong wallet icon after default wallet change in popup


Tedko (approved):
lgtm

PR #1621, opened 2020-09-26
chore: add warning in gh-actions result

septs:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message


Jack-Works:
Cool

PR #1637, opened 2020-09-30
feat: distinguish ui by platform

zhouhanseng:
fix this:
image


Jack-Works (approved):
Though I'm not happy for this change.

PR #1652, opened 2020-10-02
feat: enable hmr by default cause upstream is ready

Jack-Works:


webpack.config.ts
```diff
@@ -39,9 +39,7 @@ export default function (_argvEnv, argv: any) {
 
     const enableHMR =
         env === 'development' &&
-        !process.env.NO_HMR &&
-        // Wait for https://github.com/crimx/webpack-target-webextension/pull/4
-        false
+        !process.env.NO_HMR
```
septs, line 42:
```suggestion
        !Boolean(process.env.NO_HMR)
```

guanbinrui (approved):

PR #1655, opened 2020-10-02
fix: fit iPad extension page UI

zhouhanseng:


src/utils/theme.ts
```diff
@@ -24,6 +24,15 @@ const base: ThemeOptions = {
     typography: {
         fontFamily: getFontFamily(),
     },
+    breakpoints: {
+        values: {
+            xs: 0,
+            sm: 600,
+            md: 1112,
```
zhouhanseng, line 31: According to https://developer.apple.com/library/archive/documentation/DeviceInformation/Reference/iOSDeviceCompatibility/Displays/Displays.html, the largest width of iPad is 1112.

src/extension/options-page/DashboardDialogs/Base.tsx
```diff
@@ -149,6 +149,7 @@ const useDashboardDialogWrapperStyles = makeStyles((theme) =>
             flexDirection: 'column',
             width: (props) => (props.size === 'small' ? 280 : 440),
             padding: (props) => (props.size === 'small' ? '40px 24px !important' : '40px 36px !important'),
+            margin: '0 auto',
```
zhouhanseng, line 152: image

to

image

septs: if can, plz try use flex.

see https://css-tricks.com/snippets/css/a-guide-to-flexbox/

zhouhanseng: I know flex, but it seems not easy to do at a glance, I will check again.

zhouhanseng: I tried to set its parent to a flex box, but it doesn't looks well:

image

image

I found that its parent has property `'position': 'fix'`, then I get the answer https://stackoverflow.com/a/29036585/4803321

> Absolutely positioned boxes are **taken out of the normal flow**.

septs: OK.

septs (changes requested):
image

zhouhanseng:
Locally run `./node_modules/.bin/prettier --check ./`, it seems not caused by this PR?

webpack.config.ts
```diff
@@ -37,9 +37,7 @@ export default function (_argvEnv, argv: any) {
     const env: 'production' | 'development' = argv.mode
     const dist = env === 'production' ? src('./build') : src('./dist')
 
-    const enableHMR =
-        env === 'development' &&
-        !Boolean(process.env.NO_HMR)
+    const enableHMR = env === 'development' && !Boolean(process.env.NO_HMR)
```
zhouhanseng, line 40: Fix a prettier format issue.

Jack-Works (approved):

septs (approved):

PR #1745, opened 2020-10-19
feat: move to rush monorepo?

Jack-Works:
Try it:

```bash
npm i -g @microsoft/rush
rush update
rush build
```


Jack-Works:
merged with master, I don't know if it works.

Jack-Works:
Deferred, I tried the `rush` in a small project and found it's not easy to use. My friend told me `lerna` has this problem too. We may continue to discuss somewhere else.

Issue #1749, opened 2020-10-19
[Bug] iOS Mask Twitter hinting color bug


Tedko:
# Bug Report

<img width="325" alt="Screen Shot 2020-10-20 at 00 40 26" src="https://user-images.githubusercontent.com/6687437/96480486-de489a00-126c-11eb-9823-54786e69daec.png">

The hinting color is buggy 

### System

- [ ] Mac OS X

  - OS Version:

- [ ] Linux

  - Linux Distribution:
  - OS Version:

### Platform/Browser

- [ ] Chrome

  - Maskbook Version:
  - Browser Version:

- [ ] Firefox

  - Maskbook Version:
  - Browser Version:

- [ ] Android

  - Maskbook Version:
  - Android Version:

- [x] iOS
  - Maskbook Version:
  - iOS Version:

### Build Variant

- Where do you get Maskbook?
  - [ ] Store
  - [ ] ZIP
  - [ ] Self-Compiled
    - Build Commit: /_Optionally attach a Commit ID, if it is from an pre-release branch head_/

### Actual Behavior

/_What happened?_/

Tedko:
assign to @zhouhanseng

Tedko:
#1668

zhouhanseng:
@Tedko  

@jk234ert and I still can't find the reason.

Jack-Works:
@zhouhanseng maybe related to auto-theme detection (it might think current page is in the dark mode)

zhouhanseng:
No, the normal color of text in dark mode is different from this color.

Jack-Works:
Maybe you can print out the theme actually used to figure out why

Tedko:
> @Tedko
> 
> @jk234ert and I still can't find the reason.

can u reproduce

zhouhanseng:
> can u reproduce

no

Tedko:
> > can u reproduce
> 
> 
> 
> no

On real iOS or simulator

zhouhanseng:
simulator

Tedko:
try real iOS

Tedko:
@zhouhanseng

Jack-Works:
Is the color overwritten by a Twitter CSS? Maybe we should try to open the ShadowRoot mode on iOS again

PR #1802, opened 2020-10-28
Remove non ShadowRoot rendering mode

Jack-Works:
because it causing #1749. Need to be tested on iOS.


jk234ert:
Looks fine on iOS devices. OK to merge on my side.
BTW we still need @Tedko to re-test #1749 in the next iOS release.

Jack-Works:
#1749 is caused by host overwriting our style

Tedko:
Will test again.

PR #1871, opened 2020-11-09
refactor: migrate to new environment api

Jack-Works:


guanbinrui (approved):

Jack-Works:
> TimeoutError: Timed out after 30000 ms while trying to connect to the browser! Only Chrome at revision r756035 is guaranteed to work

PR #1878, opened 2020-11-10
feat: mobile check persona

zhouhanseng:
@jk234ert


septs:
CC @Jack-Works

Jack-Works:
We've decided to use another mechanism in the last meeting @zhouhanseng

jk234ert:
Since our injected scripts take some time (5-6 secs on outdated device models) to be executed, native could not get the actual timeframe to evaluate any function to get the value from JS side like "currently connected personas". 
IMO we still need such kind of "mobileIsMyPersonasInitialized" for native to decide the routing destination on app launch.

jk234ert (approved):

Jack-Works:
@jk234ert No it doesn't rely on "injected scripts". The background script takes the job of responsing.

PR #1889, opened 2020-11-12
chore: remove unused imports

septs:


guanbinrui (approved):

.eslintrc.json
```diff
@@ -24,6 +25,12 @@
         "eqeqeq": "error",
         "no-extra-bind": "error",
         "use-isnan": "error",
-        "no-restricted-globals": ["error", "event", "name", "length", "closed"]
+        "no-restricted-globals": ["error", "event", "name", "length", "closed"],
+        "@typescript-eslint/no-unused-vars": "off",
+        "unused-imports/no-unused-imports-ts": "error",
```
Jack-Works, line 30: No need for this rule. `noUnusedLocals` in `tsconfig` is enough

packages/maskbook/src/contracts/balance-checker/BalanceChecker.d.ts
```diff
@@ -1,11 +1,10 @@
 /* Generated by ts-generator ver. 0.0.8 */
 /* tslint:disable */
 
-import BN from 'bn.js'
```
Jack-Works, line 4: This file is auto generated. It should not be linted

PR #1894, opened 2020-11-12
feat: backport webpack 5 asset modules and worker support to 4 

Jack-Works:
# Backward port of Webpack 5 AssetModules and native worker to Webpack 4

After upgraded to webpack 5, remove this file, worker-plugin and file-loader.

- <https://webpack.js.org/guides/asset-modules/#root>
- <https://webpack.js.org/blog/2020-10-10-webpack-5-release/#asset-modules>
- <https://webpack.js.org/blog/2020-10-10-webpack-5-release/#native-worker-support>

This is not a full-implement. If you need any AssetModules feature that not supported yet, please contact @Jack-Works

## Support for AssetModules

- ✔️ `new URL("./file.png", import.meta.url)` as asset module
- ✔️ Type: asset/resource (always)
- ❌ Type: asset/inline (not supported)
- ❌ Type: asset/source (not supported)
- ❌ Loaders for asset modules
- ❌ assetModuleFilename

## Support for native Workers

- ✔️ `new *Worker(new URL("./file", import.meta.url))` as worker
- ✔️ `new *Worklet(new URL("./file", import.meta.url))` as worker
- ❌ Service Worker (it will be treaded as asset module, not a worker)


Jack-Works:
There might be some problem with `__webpack_public_path__`. I will try it later.

Jack-Works:
@guanbinrui no longer need to `import` a `png` file. This `new URL("./file.png", import.meta.url)` approach is good to prevent vender lock-down.

septs (dismissed):

septs:
```
Warning (worker-plugin): output.globalObject is set to "window". It must be set to "self" to support HMR in Workers.
```

Jack-Works:
We are using globalThis as the global object, no need to worry

PR #1899, opened 2020-11-13
chore: migrate qrcode worker

Jack-Works:


packages/maskbook/src/web-workers/OnDemandWorker.ts
```diff
@@ -0,0 +1,34 @@
...
+        console.log(`OnDemandWorker ${this.id}`, ...args)
+    }
+    protected lastUsed = Date.now()
+    protected use(onReady: () => void) {
+        this.lastUsed = Date.now()
+        this.log('kept alive')
+
+        if (this.worker) return onReady()
+        this.worker = new Worker(...this.init)
+        this.worker.addEventListener('error', (e) => this.dispatchEvent(e))
```
guanbinrui, line 21: No need for removing those listeners?

Jack-Works: After the EventTarget is being recycled, its listeners are also recycled so doesn't count to reference to the Uint8Array.

packages/maskbook/src/web-workers/OnDemandWorker.ts
```diff
@@ -0,0 +1,34 @@
...
+        this.init = init
+        this.log('created with', ...init)
+    }
+    protected log(...args: any[]) {
+        console.log(`OnDemandWorker ${this.id}`, ...args)
+    }
+    protected lastUsed = Date.now()
+    protected use(onReady: () => void) {
+        this.lastUsed = Date.now()
+        this.log('kept alive')
```
septs, line 17: ???
keep alive?

packages/maskbook/src/web-workers/OnDemandWorker.ts
```diff
@@ -0,0 +1,34 @@
...
+        this.worker = new Worker(...this.init)
+        this.worker.addEventListener('error', (e) => this.dispatchEvent(e))
+        this.worker.addEventListener('message', (e) => this.dispatchEvent(e))
+        this.worker.addEventListener('messageerror', (e) => this.dispatchEvent(e))
+        setTimeout(onReady, 50)
+    }
+    terminate() {
+        this.worker?.terminate()
+        this.worker = undefined
+        this.log('terminated')
```
septs, line 29: why not with ${this.id}?

PR #1920, opened 2020-11-18
feat: mobile launch page

zhouhanseng:
@jk234ert


packages/maskbook/src/background-service.ts
```diff
@@ -94,13 +93,6 @@ if (isEnvironment(Environment.ManifestBackground)) {
             browser.tabs.create({ url: getWelcomePageURL() })
         }
     })
-
-    contentScriptReady.then(() => {
```
zhouhanseng, line 98: Since mobile chooses which page to launch, this is no longer needed.

Jack-Works (approved):

Tedko (approved):

PR #1933, opened 2020-11-20
feat: upgrade to ts 4.1 enable new JSX transform

Jack-Works:


septs:
storybook current is unavailable. 

image

PR #1953, opened 2020-11-25
feat: support i18n namespace

Jack-Works:
Our scanning tool need to support this new pattern.


septs:
@Jack-Works I want this

```
packages/locale/en/message.json
packages/locale/en/plugin@file-service.json
```

Jack-Works:
In future plugins should host their i18n files in their own packages so I don't think it's a good idea to place them in a folder. @septs

septs:
@Jack-Works i think, we need `/docs`.

1. move `caveats.md` to `/docs/caveats.md`
2. provide i18n maintenance manual

Jack-Works:
@septs how to let the locale kit support this PR?

septs:
@Jack-Works flatten `_locales` file structure?

`_locales/*/messages.json` -> `_locales/*.json`?

Jack-Works:
Possible.

Jack-Works:
@septs this pr is depended on locale kit to be updated

Jack-Works:
@septs I thought we may find another way to do this. For example, we can learn from TypeScript repo. In TypeScript, you need to define messages in a JSON file, then run a command to generate a TS file (that not committed to the git), and then you use that file.

If we use the same approach we may detect the usage in a more efficient way, for example, each message is a property on an object and we can use the existing tooling to find how many usages on that property. If there is none, that means this key is unused. How do you think?

septs:
good idea.

Jack-Works:
This approach has a problem, we need to run a continuous FS watcher on the JSON file to generate the TS file. And if we want to analyze the unused key we need to run a language service.

PR #1959, opened 2020-11-26
feat: add gecko view message channel

Jack-Works:


jk234ert:
Tested on Android, native can not receive message from JS side.
GeckoView received connect event when manually run `let port = browser.runtime.connectNative("browser");` in console.
Then GeckoView could properly receive any message from JS by using `port.postMessage`

It seems `browser.runtime.connectNative("browser")` is not executed by default. @Jack-Works

Jack-Works:
Maybe that code didn't run... I'll try to make it more explicit.

Jack-Works:
@jk234ert try again

jk234ert:
Works properly now :)
We could remove the "hello" test now, and I will push a change to the manifest.json file to satisfy GeckoView's permissions requirement.

jk234ert:
Let me have a final test...

packages/maskbook/src/components/shared/qrcode.tsx
```diff
@@ -74,10 +74,15 @@ export function QRCode({ text, options = {}, canvasProps }: QRProps) {
...
 
-export function WKWebkitQRScanner(props: { onScan?: (val: string) => void; onQuit?: () => void }) {
+export function NativeQRScanner(props: { onScan?: (val: string) => void; onQuit?: () => void }) {
     useAsync(async () => {
         try {
-            props.onScan?.(await iOSHost.scanQRCode())
+            if (nativeAPI?.type === 'iOS') {
+                props.onScan?.(await nativeAPI.api.scanQRCode())
+            } else {
+                // TODO:
```
septs, line 83: use native qr scanner for android platform? (future)

Jack-Works: Yes, I leave it here for that purpose.

Jack-Works:
@jk234ert please notice that there might be multiple connect (generally each page including background page will have at least one connection). Please handle them correctly.
If you're responding a request or requesting some function please make sure you send the message to the correct port.

jk234ert:
> @jk234ert please notice that there might be multiple connect (generally each page including background page will have at least one connection). Please handle them correctly.
> If you're responding a request or requesting some function please make sure you send the message to the correct port.

Thanks for the notice.
I added logs where the `connectNative` method invoked, confirmed multiple connect does exist. Fortunately only the first time invoke triggers GeckoView's related callbacks. Thought GeckoView internally manages a `connected port` handler and would not creates duplicates.

jk234ert (approved):

jk234ert:
Let's create another PR for the NativeRPC's implementation

Jack-Works:
I think that doesn't correct if there is only 1 port get handled.
Imagine this scene: background send request to native (though background port) and native send request to dashboard (thought dashboard port). There're two ports comes in.

jk234ert:
> I think that doesn't correct if there is only 1 port get handled.
> Imagine this scene: background send request to native (though background port) and native send request to dashboard (thought dashboard port). There're two ports comes in.

GeckoView's connect-based messaging seems only to support a universal port. No matter how many times `browser.runtime.connectNative` only 1 port instance is sent to native, JS side has to maintain a universal port IMO

Jack-Works:
> GeckoView's connect-based messaging seems only to support a universal port. No matter how many times browser.runtime.connectNative only 1 port instance is sent to native, JS side has to maintain a universal port IMO

This doesn't seem right. Can you show the documentation? Another question, does it distinguished by the first parameter of `browser.runtime.connectNative()`? E.g. what if I call `browser.runtime.connectNative("another-string")`?

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -1,9 +1,13 @@
+import type { PersonaWithProviders } from '../../database'
 /**
  * JSON RPC calls that can be called from the native side.
  * JS = server, native = client
  */
 export interface WebviewAPIs {
     web_echo<T>(arg: T): Promise<T>
+    getUrl(path: string): string
```
Jack-Works, line 8: Please mark it as Promise<string>

packages/maskbook/src/database/Persona/types.ts
```diff
@@ -17,3 +17,15 @@ export interface Persona extends TypedOmit<PersonaRecord, 'localKey' | 'publicKe
 export interface PersonaWithPrivateKey
     extends Omit<PersonaRecord, 'privateKey'>,
         Required<Pick<PersonaRecord, 'privateKey'>> {}
+
+export interface Provider {
+    internalName: string
+    network: string
+    connected: boolean
+    userId: string | undefined
+    identifier: ProfileIdentifier | undefined
```
Jack-Works, line 26: You should not rely on `ProfileIdentifier` because it is exposed to the external. Please extract it into `network` and `userID`

Jack-Works: Oh you already have network and userId so why you need identifier?

packages/maskbook/src/database/Persona/types.ts
```diff
@@ -17,3 +17,15 @@ export interface Persona extends TypedOmit<PersonaRecord, 'localKey' | 'publicKe
...
+}
+
+export interface PersonaWithProviders extends Persona {
```
Jack-Works, line 29: Please explicitly re-describe the interface with minimal information needed instead of re-use our internal representation. If your public API depended on the internal API, there will be a refactor hazard.

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -1,9 +1,13 @@
...
+    getStorage(key: string): Promise<browser.storage.StorageObject>
```
Jack-Works, line 9: This should not be a public API. What exact information do you want to expose? Please provide an API for that purpose instead of public a method for querying our internal state.

packages/maskbook/src/utils/native-rpc/Web.ts
```diff
@@ -1,4 +1,28 @@
...
+                    const profile = profiles.find(([key, value]) => key.network === i.networkIdentifier)
+                    return {
+                        internalName: i.internalName,
+                        network: i.networkIdentifier,
+                        connected: !!profile,
+                        userId: profile?.[0].userId,
+                        identifier: profile?.[0],
+                    }
+                })
+                return { providers, ...p }
```
Jack-Works, line 25: You should not use `...p` in this case because it may extract undocumented properties. This may cause a hidden dependency (which is bad) if the consumer found some useful undocumented properties.

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -1,9 +1,13 @@
...
+    getConnectedPersona(): Promise<PersonaWithProviders[]>
```
Jack-Works, line 10:
```suggestion
    getConnectedPersonas(): Promise<PersonaWithProviders[]>
```

scripts/manifest-modifiers.ts
```diff
@@ -9,7 +9,11 @@ export function firefox(manifest: Manifest) {
 /** Geckoview is firefox with some quirks. */
 export function geckoview(manifest: Manifest) {
     firefox(manifest)
-    manifest.permissions.push('<all_urls>')
+    manifest.permissions.push(
+        'nativeMessaging', 
+        'nativeMessagingFromContent', 
+        'geckoViewAddons', 
+        '<all_urls>')
```
Jack-Works, line 16: Does this code pass prettier? 🤔

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface iOSNativeAPIs extends SharedNativeAPIs {
+    scanQRCode(): Promise<string>
+    log(...args: any[]): Promise<void>
+}
+/**
+ * JSON RPC calls that can be called if it is running on Android.
+ * JS = client, Android = server
+ */
+export interface AndroidNativeAPIs extends SharedNativeAPIs {
+    android_echo(arg: string): Promise<string>
```
Jack-Works, line 26: @jk234ert would you like to add a `scanQRCode(): Promise<string>` method for Android so we can move this API to `SharedNativeAPIs`?

packages/maskbook/src/utils/native-rpc/types.ts
zhouhanseng: `browser.runtime.getURL(path)` is not an async function, should still mark it as Promise?

packages/maskbook/src/database/Persona/types.ts
zhouhanseng: get it!

packages/maskbook/src/utils/native-rpc/types.ts
jk234ert: Yes, a native QR scan would be much better.

packages/maskbook/src/utils/native-rpc/types.ts
Jack-Works: Yes. It's always async because it is RPC. And I suggest native side to make things more clear. Expose a `getDashboardURL(): Promise<string>` instead of exposing the underlying URL detail. If so, refactoring on the URL will cause a breaking change in the native side which is unwanted.

packages/maskbook/src/utils/native-rpc/types.ts
jk234ert: App native plans to retrieve several setting values ("launchPage", etc ..) from JS side.
To ensure security JS could define a whitelist of `storage keys`, but that does not need to be present in the name of the API IMO

packages/maskbook/src/utils/native-rpc/Web.ts
zhouhanseng: ok,  mobile side only need `{network: string, connected: boolean}`

packages/maskbook/src/database/Persona/types.ts
zhouhanseng: ok,  mobile side only need `{network: string, connected: boolean}`

packages/maskbook/src/utils/native-rpc/types.ts
Jack-Works: So the implementation should not match the browser.storage.local.get. It should get value from our settings object.

zhouhanseng: done

packages/maskbook/src/utils/native-rpc/Web.ts
```diff
@@ -1,11 +1,12 @@
 import type { WebviewAPIs } from './types'
 import Services from '../../extension/service'
 import { definedSocialNetworkWorkers } from '../../social-network/worker'
+import * as settings from '../../settings/settings'
 
 export const WebviewAPI: WebviewAPIs = {
     web_echo: async (arg) => arg,
     getDashboardUrl: async (path: string) => browser.runtime.getURL(path),
-    getStorage: async (key: string) => browser.storage.local.get(key),
+    getSettings: async (key: keyof typeof settings) => settings[key].value,
```
zhouhanseng, line 9: to use type `keyof typeof settings`, I move the types of settings to`settings/types.ts`

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -0,0 +1,27 @@
+import type * as settings from '../../settings/settings'
+/**
+ * JSON RPC calls that can be called from the native side.
+ * JS = server, native = client
+ */
+export interface WebviewAPIs {
+    web_echo<T>(arg: T): Promise<T>
+    getDashboardUrl(path: string): Promise<string>
+    getSettings(key: keyof typeof settings): Promise<any>
```
Jack-Works, line 9: There are still hidden dependencies on the internal implementation. The other side is not TypeScript and does not have direct dependencies on our codebase. Please explicitly list the settings you want to expose and mark their types.

Jack-Works: I suggest it should be like:

```ts
getSettings(key: "settings-name-A"): Promise<SettingsReturnValueA>
getSettings(key: "settings-name-B"): Promise<SettingsReturnValueB>
```

use overload to mark what settings can be used and it will easier to read by @jk234ert

packages/maskbook/src/settings/types.ts
```diff
@@ -0,0 +1,34 @@
...
+    username?: string
+    /** The WIP step */
+    status?: SetupGuideStep
+}
+
+export enum LaunchPage {
+    facebook = 'facebook',
+    twitter = 'twitter',
+    dashboard = 'dashboard',
+}
```
Jack-Works, line 34: It seems like this file is exposed as part of public API to the native side. Please leave a comment to notice that if you make an incompatible change in this file, it will break the API.

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -0,0 +1,27 @@
...
+    getConnectedPersonas(): Promise<{ network: string; connected: boolean }[][]>
```
Jack-Works, line 10: Hmm, why it is `{...}[][]` instead of `{...}[]`?

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -0,0 +1,27 @@
+import type * as settings from '../../settings/settings'
+/**
+ * JSON RPC calls that can be called from the native side.
+ * JS = server, native = client
+ */
+export interface WebviewAPIs {
+    web_echo<T>(arg: T): Promise<T>
+    getDashboardUrl(path: string): Promise<string>
```
Jack-Works, line 8: Please explicitly mark available "intent" instead of receiving a "path". BTW it should be named as `getDashboardURL`.

The API should looks like:

```ts
getDashboardURL(intent: "wallets" | "personas" | "other pages..."): Promise<string>
```

packages/maskbook/src/utils/native-rpc/types.ts
zhouhanseng: The api returns `[[{network: 'facebook.com', connected: false}, {network: 'twitter.com', connected: true}],[//...],[//...]] `

packages/maskbook/src/utils/native-rpc/types.ts
jk234ert: @zhouhanseng For now please just remove the parameter, native only needs the dashboard root url

packages/maskbook/src/utils/native-rpc/types.ts
Jack-Works: So what difference between `result[0]` and `result[1]`? Are they belong to different personas? If so why there isn't any thing to mark it?

zhouhanseng: They belong to different personas. @jk234ert told me it is okay even there's nothing to mark it.

packages/maskbook/src/utils/native-rpc/types.ts
zhouhanseng: ok

zhouhanseng: done

packages/maskbook/src/settings/types.ts
zhouhanseng: done

packages/maskbook/src/utils/native-rpc/types.ts
zhouhanseng: ok

packages/maskbook/src/utils/native-rpc/types.ts
Jack-Works: If the API is ready please cc me @jk234ert

packages/maskbook/src/settings/types.ts
```diff
@@ -0,0 +1,38 @@
+// Hack: this file is exposed as part of public API to the native side,
```
Jack-Works, line 1: It's not a `HACK`, it is a formal document.

packages/maskbook/src/settings/types.ts
```diff
@@ -0,0 +1,38 @@
...
+export type SetupGuideCrossContextStatus = {
+    /** The persona to be connected */
+    persona?: string
+    /** The user name given by user */
+    username?: string
+    /** The WIP step */
+    status?: SetupGuideStep
+}
+
+export enum LaunchPage {
```
Jack-Works, line 32: Maybe mark this enum with TSDoc comment `@public`?

Jack-Works:
```suggestion
/** @public */
export enum LaunchPage {
```

packages/maskbook/src/settings/types.ts
Jack-Works:
```suggestion
// WARNING: Types with @public mark is exposed as public API to the native side,
```

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -0,0 +1,27 @@
+import type { LaunchPage, NativeAPISettingsName } from '../../settings/types'
+/**
+ * JSON RPC calls that can be called from the native side.
+ * JS = server, native = client
+ */
+export interface WebviewAPIs {
+    web_echo<T>(arg: T): Promise<T>
+    getDashboardURL(): Promise<string>
+    getSettings(key: NativeAPISettingsName): Promise<LaunchPage>
```
Jack-Works, line 9:
```suggestion
    /** Definition of LaunchPage see https://github.com/DimensionDev/Maskbook/blob/master/packages/maskbook/src/settings/types.ts */
    getSettings(key: 'launchPageSettings'): Promise<LaunchPage>
```

packages/maskbook/src/utils/native-rpc/types.ts
jk234ert: Separate issue created for the Android native scan QR code feature:
#2039 
@Jack-Works this needs some work on the Android native side, as an enhancement.

packages/maskbook/src/utils/native-rpc/Web.ts
```diff
@@ -0,0 +1,27 @@
+import type { WebviewAPIs } from './types'
+import Services from '../../extension/service'
+import { definedSocialNetworkWorkers } from '../../social-network/worker'
+import * as settings from '../../settings/settings'
+import type { NativeAPISettingsName } from '../../settings/types'
+
+export const WebviewAPI: WebviewAPIs = {
+    web_echo: async (arg) => arg,
+    getDashboardURL: async () => browser.runtime.getURL(''),
+    getSettings: async (key: NativeAPISettingsName) => settings[key].value,
```
Jack-Works, line 10: You don't need to type the `key` here. Please remove type `NativeAPISettingsName`. It can be inferred from `WebviewAPIs` IMO.

zhouhanseng: yes it is.

zhouhanseng: No, the editor complains that key has an any type.

zhouhanseng: yes it is.

Jack-Works:
the current changes look good to me

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -8,7 +8,7 @@ export interface WebviewAPIs {
     getDashboardURL(): Promise<string>
     /** Definition of LaunchPage see https://github.com/DimensionDev/Maskbook/blob/master/packages/maskbook/src/settings/types.ts */
     getSettings(key: 'launchPageSettings'): Promise<LaunchPage>
-    getConnectedPersonas(): Promise<{ network: string; connected: boolean }[][]>
+    getConnectedPersonas(): Promise<{ personas: { network: string; connected: boolean }[][] }>
```
zhouhanseng, line 11: Android can't serialize array.

Jack-Works: interesting. is there any upstream bug link?

jk234ert: It's not `cannot serialize array`, it's GeckoView fails to json-fy the output of `getConnectedPersonas`
@zhouhanseng I tried using `Json.stringfy()` to encode the output and it works on Android.
Guess we do not need to add this key, just stringfy the result of this method?

packages/maskbook/src/utils/native-rpc/Web.ts
```diff
@@ -0,0 +1,28 @@
+import type { WebviewAPIs } from './types'
+import Services from '../../extension/service'
+import { definedSocialNetworkWorkers } from '../../social-network/worker'
+import * as settings from '../../settings/settings'
+
+export const WebviewAPI: WebviewAPIs = {
+    web_echo: async (arg) => arg,
+    getDashboardURL: async () => browser.runtime.getURL(''),
```
jk234ert, line 8: How about 
```suggestion
    getDashboardURL: async () => browser.runtime.getURL('index.html'),
```
App needs to load the entry page of dashboard, and the path `index.html` is configured by JS.
Or we could use a predefined path constant instead here

jk234ert (changes requested):

packages/maskbook/src/utils/native-rpc/types.ts
Jack-Works: why failed to JSON-ify? Is it containing circular reference or it's a bug of GeckoView?

Jack-Works: Or it's an exotic object (For example, a Proxy, DOM object, Date object, ... that created by the JS engine) / proptotyped object (a class) that cannot be serialized?

jk234ert: GeckoView simply logs the error message "fail to dispatch message" on the JS side and hides the underneath error root cause.

Tested on GeckoView debug console as following:
1. `let aport = browser.runtime.connectNative("browser"); ` OK
2. `aport.postMessage([])` OK
3. `aport.postMessage([[]])` `**Error dispatching GeckoView:WebExtension:PortMessage Messaging.jsm:76:24
Invalid event data property data**`

Seems GeckoView cannot JSON-ify an object with nested array

jk234ert (dismissed):

packages/maskbook/src/utils/native-rpc/types.ts
Jack-Works: 😂 @jk234ert please report it to Firefox and leave a link here

packages/maskbook/src/utils/native-rpc/types.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * JSON RPC calls that can be called from the native side.
+ * JS = server, native = client
+ */
+export interface WebviewAPIs {
+    web_echo<T>(arg: T): Promise<T>
+    getDashboardURL(): Promise<string>
+    /** Definition of LaunchPage see https://github.com/DimensionDev/Maskbook/blob/master/packages/maskbook/src/settings/types.ts */
+    getSettings(key: 'launchPageSettings'): Promise<LaunchPage>
+    getConnectedPersonas(): Promise<string>
```
Jack-Works, line 11: We should describe the return value in the comment as document.

Jack-Works:
rebased to master branch

Jack-Works:
the current changes LGTM, please use squash and merge. please review @septs @guanbinrui

septs (approved):

packages/maskbook/src/utils/native-rpc/types.ts
jk234ert: Bug reported to Mozilla, [link](https://bugzilla.mozilla.org/show_bug.cgi?id=1681638)

PR #1981, opened 2020-12-01
feat: remove copy to clipboard behavior

septs:


Jack-Works:
If it's no longer copying it's elements, why should it be named the Copyable?

Issue #2015, opened 2020-12-03
[Bug] Copied Notifcation overlay-ed by Twitter DM UX


Tedko:
## Bug Report

<img width="1916" alt="Screen Shot 2020-12-04 at 00 28 06" src="https://user-images.githubusercontent.com/6687437/101062657-e1090f80-35cc-11eb-9218-c087d0af5e78.png">

How to reproduce:

Search $uni  or $eth on twitter; 
click to copy （see down here)

<img width="1939" alt="Screen Shot 2020-12-04 at 01 10 34" src="https://user-images.githubusercontent.com/6687437/101063507-d56a1880-35cd-11eb-8138-3429994b1045.png">

Then you will see the copy notif is overlay-ed by twit dm ux

### System

- [x] Mac OS X

  - OS Version:

### Platform/Browser

- [ ] Chrome

  - Maskbook Version:
  - Browser Version:

- [ ] Firefox

  - Maskbook Version:
  - Browser Version:

- [ ] Android

  - Maskbook Version:
  - Android Version:

- [ ] iOS
  - Maskbook Version:
  - iOS Version:

### Build Variant

- Where do you get Maskbook?
  - [x] Store

### Actual Behavior

/_What happened?_/

septs:
#1975 same question

Tedko:
Ok. Let’s do a quick fix. Move it from right to left? (Like the img payload
notify

septs:
CC @Jack-Works

septs:
Related code: https://github.com/DimensionDev/Maskbook/blob/ab8a1e6ce3394794cc2271a05aeda56d0ec8b6b0/packages/maskbook/src/utils/shadow-root/renderInShadowRoot.tsx#L240-L244

Jack-Works:
@septs OK to change elsewhere. Or you can wrap another SnackbarProvider

Tedko:
Let's do a quick fix buy change it to right side. Or if u guys can make the DM not overlay it. Whichever faster

Tedko:
Cc @Jack-Works @septs

PR #2064, opened 2020-12-10
fix: move snackbar to top right close gh-2015

Jack-Works:


septs (commented):
Mobile only? need test?

jk234ert:
This fix should be desktop only

jk234ert:
Tested on Chrome & Firefox
Snackbar displays at the left bottom corner, is this by design? @Jack-Works 

Snipaste_2020-12-10_16-31-54

Tedko:
> Tested on Chrome & Firefox
> Snackbar displays at the left bottom corner, is this by design? @Jack-Works
> 
> Snipaste_2020-12-10_16-31-54

this part is ok

Tedko:
@jk234ert  this is for https://github.com/DimensionDev/Maskbook/issues/2015

should be only related w/ desktop

Tedko:
btw. should we add this for mobile as well? For mobile #2015 's copy button will not have any notif:
IMG_8865



We need sth like this:

IMG_8866

Tedko (approved):

PR #2082, opened 2020-12-11
feat: non-production use nightly logo

septs:


guanbinrui (approved):

packages/maskbook/src/extension/popup-page/UI.tsx
```diff
@@ -103,11 +103,20 @@ function PopupUI() {
         window.close()
     }, [setSelectProviderDailogOpen])
 
+    const Trademark = memo(() => {
+        if (ui.networkIdentifier !== 'localhost') {
```
Jack-Works, line 107: No need to use memo here. Cause ui.networkIdentifier might change (maybe?)

PR #2084, opened 2020-12-11
fix: remove `grid display` of containers

jk234ert:


zhouhanseng (dismissed):

Jack-Works (approved):

Jack-Works (dismissed):

Jack-Works (approved):

Jack-Works (dismissed):

packages/maskbook/src/components/InjectedComponents/PostDialog.tsx
```diff
@@ -217,7 +218,7 @@ export function PostDialogUI(props: PostDialogUIProps) {
                             }}>
                             <ClickableChip
                                 checked={props.imagePayload}
-                                label={t('post_dialog__image_payload')}
+                                label={t('post_dialog__image_payload') + Flags.has_no_browser_tab_ui ? `(Beta)` : ''}
```
Jack-Works, line 221: This will also add `beta` to Android App. Your commit writes `add (Beta) to image payload option on iOS`. Is that expected?

jk234ert: Thought Android GeckoView has browser tab api.
Should I use `support_native_injected_script_declaration ` instead?

Jack-Works: No. What exact description do you want? On iOS or all Apps (iOS and Android)?

jk234ert: Only on iOS, but it seems we do not export the const `is_iOSApp ` from our `flags.ts`

septs: https://github.com/DimensionDev/Maskbook/blob/e46948584980c7b3380cd53806496a0eda9233b3/packages/maskbook/src/components/InjectedComponents/PostDialog.tsx#L134-L142

Jack-Works: @septs this is not a plugin

Jack-Works: @jk234ert so why it's beta on iOS but not on android?

jk234ert: Confirmed with design, should add the `(Beta)`  sub on both mobile platforms

Jack-Works: Hold on let me change it

PR #2126, opened 2020-12-17
feat: improve workflows

septs:


.github/workflows/linters.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   linters:
     runs-on: ubuntu-20.04
-    continue-on-error: false
+    continue-on-error: true
```
Jack-Works, line 12: What does it mean🤔

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
+name: Build
+
+on:
+  push:
+    branches: [master, released]
```
Jack-Works, line 5: No need for released branch

.github/workflows/linters.yml
septs: https://github.com/DimensionDev/Maskbook/runs/1569259125?check_suite_focus=true

more problems can be found

e.q: code style and type error

Jack-Works (approved):

PR #1980, opened 2020-12-01
feat: add new package components for theme design

Jack-Works:

PR #2133, opened 2020-12-22
feat: enable project reference

Jack-Works:
this is required (IMO) for #1980 cc @septs @guanbinrui plz review


.github/workflows/linters.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - run: sudo npm install --global npm@7
       - run: npm ci
       - run: npx prettier --check .
-      - run: npx tsc --noEmit
+      - run: npx tsc -b
```
septs, line 27: why?

septs (approved):

.github/workflows/linters.yml
Jack-Works: Because `--noEmit` doesn't work in project reference

Issue #2137, opened 2020-12-23
[Bug] click an option hundreds of times and after refresh I couldn't get voting result data.


xichi:
### System

- [x] Windows

  - OS Version:

- [ ] Mac OS X

  - OS Version:

- [ ] Linux

  - Linux Distribution:
  - OS Version:

### Platform/Browser

- [x] Chrome

  - Maskbook Version:
  - Browser Version:

- [ ] Firefox

  - Maskbook Version:
  - Browser Version:

- [ ] Android

  - Maskbook Version:
  - Android Version:

- [ ] iOS
  - Maskbook Version:
  - iOS Version:

### Build Variant

- Where do you get Maskbook?
  - [ ] Store
  - [ ] ZIP
  - [x] Self-Compiled
    - Build Commit: 7a5f22ced02780670dad744911bd84047c5ebe3c
    
### Actual Behavior

One user tells that he's able to click 100 times of one option and after refresh the option (100times) disappreaed.
I tried to reproduce the bug, found that `PluginPollRPC.vote` returns an error message:`RangeError: Maximum call stack size exceeded(-1)`.
![bug](https://user-images.githubusercontent.com/48639990/102968807-7f82f380-452f-11eb-9151-6567e02a919a.png).
 After refreshing the page，`PluginPollRPC.getPollByKey`  will aslo return the same the error message.So, we won't get the data of the voting result, it will be shown like this (default data):
![poll](https://user-images.githubusercontent.com/48639990/102970289-354f4180-4532-11eb-9947-dc7ddcd175ea.png)
Data can be returned after reloading extensions.I don't know why after refreshing the page, it will also return `RangeError: Maximum call stack size exceeded(-1)`. I guess it's related to [async-call-rpc](https://github.com/Jack-Works/async-call-rpc).

### Solution
I try to avoid stack overflow errors：
1. Bind the debounce function to `PluginPollRPC.vote`，avoiding continuous click.
2. Add a new PollStatus: Error.
![error](https://user-images.githubusercontent.com/48639990/102973980-b27db500-4538-11eb-87e1-cc496c148e3c.png)

1999-XiChi:
PR: https://github.com/DimensionDev/Maskbook/pull/2138

1999-XiChi:
Is there any better solutions? @Jack-Works

PR #2138, opened 2020-12-23
fix: add an error status message when vote fails

xichi:
closes https://github.com/DimensionDev/Maskbook/issues/2137


packages/maskbook/src/_locales/zh/messages.json
```diff
@@ -392,5 +392,6 @@
     "plugin_poll_status_closed": "投票已截止",
     "plugin_poll_status_voting": "投票中",
     "plugin_poll_status_voted": "已投票.",
+    "plugin_poll_status_error": "蔙誤!",
```
septs, line 395:
```suggestion
    "plugin_poll_status_error": "錯誤!",
```

packages/maskbook/src/plugins/Polls/UI/Polls.tsx
```diff
@@ -126,9 +135,12 @@ export function PollCardUI(props: PollCardProps) {
                     <ListItem
                         className={classes.option}
                         key={index}
-                        onClick={() => {
-                            vote?.(poll, index)
-                        }}>
+                        onClick={debounce(
```
Jack-Works, line 138: IMO it will create a new "debounced" button every time so it will not work

Tedko: @1999-XiChi cc

septs:
outdated

PR #2174, opened 2021-01-04
fix: contacts cannot be displayed in mobile

lelenei:
Set the height of the phone according to CSS media max-width

### DIFF
--- a/packages/maskbook/src/extension/options-page/DashboardRouters/Contacts.tsx
+++ b/packages/maskbook/src/extension/options-page/DashboardRouters/Contacts.tsx
@@ -35,7 +35,33 @@ const useStyles = makeStyles((theme) =>
                 marginLeft: theme.spacing(-2),
                 marginRight: theme.spacing(-2),
             },
+            display: 'flex',
         },
+        fixed: {
+            height: '100%',
+            width: '100%',
+            ['@media (max-width: 320px)']: {
+                height: 400,
+
+            },
+            ['@media (max-width: 360px)']: {
+                height: 491,
+            },
+            ['@media (max-width: 375px)']: {
+                height: 581,
+
+            },
+            ['@media (max-width: 411px)']: {
+                height: 602,
+            },
+            ['@media (max-width: 414px)']: {
+                height: 607,
+
+            },
+            ['@media (max-width: 540px)'] : {
+                height: 581,
+            },
+        }
     }),
 )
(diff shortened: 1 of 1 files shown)


packages/maskbook/src/extension/options-page/DashboardRouters/Contacts.tsx
```diff
@@ -151,32 +177,34 @@ export default function DashboardContactsRouter() {
...
+                                    if (isPagePending) return
+                                    if (data.visibleStopIndex === data.overscanStopIndex) nextPage()
+                                }}
+                                itemSize={64}
+                                itemCount={items.length}
+                                {...sizeProps}>
+                                {({ index, style }) =>
+                                    items[index] ? (
+                                        <ContactLine
+                                            style={style as any}
```
septs, line 196: ??????

packages/maskbook/src/extension/options-page/DashboardRouters/Contacts.tsx
```diff
@@ -35,7 +35,33 @@ const useStyles = makeStyles((theme) =>
                 marginLeft: theme.spacing(-2),
                 marginRight: theme.spacing(-2),
             },
+            display: 'flex',
         },
+        fixed: {
+            height: '100%',
+            width: '100%',
+            ['@media (max-width: 320px)']: {
```
Jack-Works, line 43: hmmm, this section seems weird cc @lelenei can you explain?

lelenei: The height of the mobile phone screen cannot be the maximum, so the height of the mobile phone is set according to the width

lelenei: Height: '100%' should be deleted

jk234ert: On mobile (in fact, when the viewport's width <= 600), the height of the contacts list on the contact page, is computed as 0.

lelenei: I think this page is not adapted for mobile phones, so I just manually calculated the required height.Of course this is incomplete.

Jack-Works:
let me create a new fix for this

PR #2176, opened 2021-01-04
feat: add skynet service provider on file service plugin

developerfred:
## Todo List
I added a new tab behavior for the user to choose which service he wants to upload

I added a list of portals that accept file uploads in the constants file
```javascript 
SIASKY_URL, LUXOR_URL, TUTEMWESI_URL, VAULT_URL, SIALOOP_URL, SIACDN_URL
```

Skynet SDK still does not support sending encrypted files.

```typescript 
// NOTE: this feature has not yet been implemented for this SDK.
export async function uploadEncryption(file: any, skykeyName: any) {
    const portalsUrls = [SIASKY_URL, LUXOR_URL, TUTEMWESI_URL, VAULT_URL, SIALOOP_URL, SIACDN_URL]
    const portalTouse = portalsUrls[Math.floor(Math.random() * portalsUrls.length)]
    const skylink = await client.uploadFile(file, portalTouse, skykeyName)

    return skylink
}```

PR #2239, opened 2021-01-13
feat!: upgrade version of file service metadata

Jack-Works:
Now FileService will emit v2 FileInfo so the older version of Mask Network needs to upgrade to see the content of v2.

I didn't test it. @septs 

It's a necessary part of #2176 . @developerfred may want to take a look about this PR

### DIFF
--- a/packages/maskbook/src/plugins/FileService/Preview.tsx
+++ b/packages/maskbook/src/plugins/FileService/Preview.tsx
@@ -47,7 +47,7 @@ const useStyles = makeStyles((theme) => ({
     },
 }))
 
-export const Preview: React.FC<Props> = ({ info }) => {
+export function Preview({ info }: { info: FileInfo }) {
     const { t } = useI18N()
     const classes = useStyles()
     const fileKey = info.key ? (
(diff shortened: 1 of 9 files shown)


developerfred:
> Now FileService will emit v2 FileInfo so the older version of Mask Network needs to upgrade to see the content of v2.
> 
> I didn't test it. @septs
> 
> It's a necessary part of #2176 . @developerfred may want to take a look about this PR

Thanks!

Jack-Works:
@septs plz checkout this

packages/maskbook/src/plugins/FileService/Preview.tsx
```diff
@@ -47,7 +47,7 @@ const useStyles = makeStyles((theme) => ({
     },
 }))
 
-export const Preview: React.FC<Props> = ({ info }) => {
+export function Preview({ info }: { info: FileInfo }) {
```
septs, line 50: revert it.

septs (approved):

packages/maskbook/src/plugins/FileService/Preview.tsx
Jack-Works: the function is better if possible to use. it is easier to analyze for the tools and can avoid TDZ problem

septs: TDZ?

Jack-Works: Temporary dead zone.

Jack-Works:
@developerfred hi please pick or rebase this PR, this is much friendly to add a new file provider type

Issue #2261, opened 2021-01-20
[Bug] Add loading indicator and "empty assets" messages to Dashboard Wallet page while fetching wallet assets


jk234ert:
### System

- [ ] Windows

  - OS Version:

- [ ] Mac OS X

  - OS Version:

- [ ] Linux

  - Linux Distribution:
  - OS Version:

### Platform/Browser

- [ ] Chrome

  - Mask Version:
  - Browser Version:

- [ ] Firefox

  - Mask Version:
  - Browser Version:

- [ ] Android

  - Mask Version:
  - Android Version:

- [ ] iOS
  - Mask Version:
  - iOS Version:

### Build Variant

- Where do you get Mask?
  - [x] Store
  - [ ] ZIP
  - [ ] Self-Compiled
    - Build Commit: /_Optionally attach a Commit ID, if it is from an pre-release branch head_/

## Bug Info
In Mask Network dashboard wallet page, since we need to request the wallet assets information by network requests, this page should display a loading indicator before network requests finish. We also need messages like "No collectibles/No Assets Found" when the current wallet is empty.

### Actual Behavior
The wallet page just presents a blank view before network requests finish.
<img width="1099" alt="Snipaste_2021-01-20_09-33-34" src="https://user-images.githubusercontent.com/7972003/105114795-97d03900-5b02-11eb-98e3-c3dbfc153800.png">

The wallet page only displays a blank view when there are no collectibles in the wallet.
<img width="1113" alt="Snipaste_2021-01-20_09-34-33" src="https://user-images.githubusercontent.com/7972003/105114907-cfd77c00-5b02-11eb-86e5-afc7e80f9f0d.png">

### Expected Behavior
1. Display "Loading..." before network requests finish on the wallet page, for both "Token" and "Collectibles" tabs. Please refer to the style in the "Add Token" dialog.
2. Display "No Data" when there are no assets/collectibles found in the wallet. Please refer to the style in the "Activity" dialog.

Tedko:
https://github.com/DimensionDev/Maskbook/issues/2117

Better not use `empty assets`

on layer 1 of eth we should better always display eth since that the gas of any tx etc. Maybe add DAI as a default currency (stable coin) even if the balance is 0

Tedko:
cc @jk234ert

nuanyang233:
@jk234ert There are two album in collectibles.So we need show duplicate indicator and empty assets?

nuanyang233:
[...]

PR #2266, opened 2021-01-20
fix: improve wallets ui

lelenei:
fix: #2117
fix: #2261


Jack-Works:
about #2261, we have [concurrent mode](https://reactjs.org/docs/concurrent-mode-intro.html) enabled. So you don't need to check the "loading" state manually. An alternative way is: trigger the suspended state and let the `<Suspense>` fallback show the loading state.

zhouhanseng:
Here's an example of Integrating with suspense: https://github.com/DimensionDev/Maskbook/pull/2221

packages/maskbook/src/web3/hooks/useAssetsDetailedCallback.ts
```diff
@@ -27,6 +27,7 @@ export function useAssetsDetailedCallback(tokens: (EtherTokenDetailed | ERC20Tok
 
     // filter out tokens in blacklist
     return {
+        loading: loading && ethloading,
```
guanbinrui, line 30:
```suggestion
        loading: loading && loadingEtherToken,
```

packages/maskbook/src/extension/options-page/DashboardComponents/WalletAssetsTable.tsx
```diff
@@ -73,15 +76,30 @@ export interface WalletAssetsTableProps extends withClasses<KeysInferFromUseStyl
...
+        setViewLength(
+            detailedTokens.length > MAX_TOKENS_LENGTH
+                ? haveETH
+                    ? MAX_TOKENS_LENGTH - 1
+                    : MAX_TOKENS_LENGTH
+                : detailedTokens.length,
+        )
+    }, [detailedTokens.length, haveETH])
+
+    if (!chainIdValid) return null
```
guanbinrui, line 102:
```suggestion
```

packages/maskbook/src/extension/options-page/DashboardComponents/WalletContent.tsx
```diff
@@ -113,6 +124,9 @@ export const WalletContent = forwardRef<HTMLDivElement, WalletContentProps>(func
     }, [])
     //#endregion
 
+    //#collectibles loading
```
guanbinrui, line 127: Maybe move `useERC721TokenDetailed` to this component, life will be simpler.

nuanyang233: The best way is trigger suspense fallback loading..so we have a library to turn the async function to suspend? like 
 `getERC721TokenDetailed`

guanbinrui: Sounds great! Let's rewrite those utility hooks to suspends.

lelenei: ok

zhouhanseng: Since `useERC721TokenDetailed` uses other hooks, it still need to be named `useERC721TokenDetailed`.

guanbinrui, line 127: Use `#region` and `#endregion`.

packages/maskbook/src/web3/hooks/useERC721TokenDetailed.ts
```diff
@@ -5,7 +5,7 @@ import { EthereumTokenType, ERC721TokenDetailed } from '../types'
 import { useChainId } from './useChainState'
 import { formatChecksumAddress } from '../../plugins/Wallet/formatter'
 import { useERC721TokenContract } from '../contracts/useERC721TokenContract'
-import { useSingleContractMultipleData } from './useMulticall'
+import { MulticalStateType, useSingleContractMultipleData } from './useMulticall'
```
Jack-Works, line 8: Unused import?

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,42 @@
...
+
+let token_: ERC721TokenDetailed | undefined = undefined
+async function ERC721TokenDetailed_(address: string, token?: Partial<ERC721TokenDetailed>) {
+    const chainId = useChainId()
+    const erc721TokenContract = useERC721TokenContract(address)
+
+    // compose calls
+    const { names, callDatas } = useMemo(
+        () => ({
+            names: ['name', 'symbol', 'baseURI'] as 'name'[],
```
Jack-Works, line 16: this type cast looks strange

packages/maskbook/src/extension/options-page/DashboardComponents/WalletAssetsTable.tsx
```diff
@@ -76,16 +79,28 @@ export interface WalletAssetsTableProps extends withClasses<KeysInferFromUseStyl
...
 
-    if (!detailedTokens.length) return null
+    const chainIdValid = useChainIdValid()
+
+    const haveETH = detailedTokens.some((x) => isETH(x.token.address))
+    const [viewLength, setViewLength] = useState(haveETH ? MAX_TOKENS_LENGTH - 1 : MAX_TOKENS_LENGTH)
+
+    useEffect(() => {
+        setViewLength(
+            detailedTokens.length > MAX_TOKENS_LENGTH
```
Jack-Works, line 97: It looks like a computed value, please don't sync it with an effect, compute it directly (or if you have special reason please comment it)

packages/maskbook/src/extension/options-page/DashboardComponents/WalletAssetsTable.tsx
```diff
@@ -157,42 +172,63 @@ export function WalletAssetsTable(props: WalletAssetsTableProps) {
...
         </div>
     )
 
+    const filterETH = (a: AssetDetailed) => isETH(a.token.address)
+    const excludeETH = (a: AssetDetailed) => !isETH(a.token.address)
+    const excludeZero = (a: AssetDetailed) => new BigNumber(a.value?.[CurrencyType.USD] || '0').isGreaterThan(0)
+    const filter = (a: AssetDetailed) =>
+        Number(price) !== 0 ? new BigNumber(a.value?.[CurrencyType.USD] || '0').isGreaterThan(price) : true
+    const sort = (a: AssetDetailed, b: AssetDetailed): number =>
+        new BigNumber(b.value?.[CurrencyType.USD] ?? 0).minus(a.value?.[CurrencyType.USD] ?? 0).toNumber()
```
Jack-Works, line 190: hoist those 5 functions to the top level if possible

.vscode/settings.json
```diff
@@ -19,5 +19,6 @@
             "url": "/.vscode/schema/locale.json"
         }
     ],
-    "cSpell.words": ["Reificated"]
+    "cSpell.words": ["Reificated"],
+    "i18n-ally.localesPaths": "packages/maskbook/src/plugins/Trader/messages"
```
Jack-Works, line 23: revert this change plz

packages/maskbook/src/web3/hooks/useERC721TokenDetailed.ts
```diff
@@ -12,13 +12,8 @@ export function useERC721TokenDetailed(address: string, token?: Partial<ERC721To
...
 
     // compose calls
-    const { names, callDatas } = useMemo(
-        () => ({
-            names: ['name', 'symbol', 'baseURI'] as 'name'[],
-            callDatas: new Array(3).fill([]),
-        }),
-        [],
-    )
+    const names = ['name', 'symbol', 'baseURI'] as 'name'[]
```
Jack-Works, line 15: why cast it as 'name'[]?

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
Jack-Works: why cast it as 'name'[]?

.vscode/settings.json
lelenei: This is automatically modified by vscode

lelenei: this is modified by prettier

Jack-Works: no it's not, it's added by i18n-ally vscode plugin

.vscode/settings.json
```diff
@@ -19,5 +19,10 @@
             "url": "/.vscode/schema/locale.json"
         }
     ],
-    "cSpell.words": ["Reificated"]
+    "cSpell.words": ["Reificated"],
+    "i18n-ally.localesPaths": [
+        "packages/dashboard/src/locales",
+        "packages/theme/src/locales",
+        "packages/maskbook/src/plugins/Trader/messages"
```
Jack-Works, line 26: 🤔or maybe you should add all locale paths. but I doubt if it works with monorepo well.

packages/maskbook/src/extension/options-page/DashboardComponents/WalletAssetsTable.tsx
```diff
@@ -73,19 +76,31 @@ export interface WalletAssetsTableProps extends withClasses<KeysInferFromUseStyl
...
 
     const classes = useStylesExtends(useStyles(), props)
     const LABELS = [t('wallet_assets'), t('wallet_price'), t('wallet_balance'), t('wallet_value'), ''] as const
 
-    const [viewLength, setViewLength] = useState(MAX_TOKENS_LENGTH)
     const [more, setMore] = useState(false)
     const [price, setPrice] = useState(MIN_VALUE)
 
-    if (!detailedTokens.length) return null
+    const haveETH = detailedTokens.some((x) => isETH(x.token.address))
```
Jack-Works, line 96:
```suggestion
    const haveETH = detailedTokens.some(filterETH)
```

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
Jack-Works: request a change here

.vscode/settings.json
lelenei: I upgrade i18n plugin in vscode

.vscode/settings.json
lelenei: Yes!

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,38 @@
...
+        name: name ?? token?.name ?? '',
+        symbol: symbol ?? token?.symbol ?? '',
+        baseURI: baseURI ?? token?.baseURI ?? '',
+    } as ERC721TokenDetailed
+
+    return
+}
+
+export function getERC721TokenDetailed(address: string, token?: Partial<ERC721TokenDetailed>) {
+    if (!token_) throw ERC721TokenDetailed_(address)
```
guanbinrui, line 36: Suppose we are using this function in different places on the same page. Will this function always return the result of the first call? Cause you are using a local variable to store the fetched token.

packages/maskbook/src/extension/options-page/DashboardComponents/WalletAssetsTable.tsx
```diff
@@ -157,42 +172,58 @@ export function WalletAssetsTable(props: WalletAssetsTableProps) {
...
                 setPrice(more ? MIN_VALUE : 0)
             }}>
-            <IconButton>{more ? <ExpandLessIcon /> : <ExpandMoreIcon />}</IconButton>
+            {detailedTokens.length > MAX_TOKENS_LENGTH ? (
+                <IconButton>{more ? <ExpandLessIcon /> : <ExpandMoreIcon />}</IconButton>
+            ) : null}
         </div>
     )
 
+    const filter = (a: AssetDetailed) =>
```
guanbinrui, line 184: Please use a proper name instead.

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
lelenei: Token is unchanged and can be used directly in other places

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,38 @@
...
+import { formatChecksumAddress } from '../../plugins/Wallet/formatter'
+import { useERC721TokenContract } from '../contracts/useERC721TokenContract'
+import { useChainId } from '../hooks/useChainState'
+import { useSingleContractMultipleData } from '../hooks/useMulticall'
+import { ERC721TokenDetailed, EthereumTokenType } from '../types'
+
+const cache = new Map<string, ERC721TokenDetailed | undefined>()
+async function ERC721TokenDetailed_(address: string, token?: Partial<ERC721TokenDetailed>) {
+    const chainId = useChainId()
+    const erc721TokenContract = useERC721TokenContract(address)
```
Jack-Works, line 11: IMO this is not correct. You're violating React hooks rules. Because this function is conditionally called. If you're intended to do this, please call those hooks in `getERC721TokenDetailed` unconditionally and rename it to `useERC721TokenDetailed`.

guanbinrui:
Hi @lelenei Could you please paste some screenshots here?

lelenei:
image

packages/maskbook/src/extension/options-page/DashboardComponents/WalletContent.tsx
```diff
@@ -187,9 +198,17 @@ export const WalletContent = forwardRef<HTMLDivElement, WalletContentProps>(func
...
-                {Flags.COTM_enabled && tabIndex === 1 ? <COTM_TokenAlbum /> : null}
-                {Flags.election2020_enabled && tabIndex === 1 ? <ElectionTokenAlbum /> : null}
+                ) : (
+                    <Suspense
+                        fallback={
+                            <Typography variant="body1" color="textSecondary">
+                                {t('loading')}
+                            </Typography>
+                        }>
+                        {Flags.COTM_enabled ? <COTM_TokenAlbum /> : null}
```
Jack-Works, line 208: Each plugin should have it's own Suspense

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,44 @@
...
+    chainId: ChainId,
+    contract: Contract,
+    address: string,
+    token?: Partial<ERC721TokenDetailed>,
+) {
+    // compose calls
+    const names = ['name', 'symbol', 'baseURI'] as (keyof ERC721['methods'])[]
+    const callDatas = new Array(3).fill([])
+
+    const [results, calls, _, callback] = useSingleContractMultipleData(contract, names, callDatas)
```
Jack-Works, line 20: No hooks in this function, please

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,44 @@
...
+    const token_ = {
+        type: EthereumTokenType.ERC721,
+        address: formatChecksumAddress(address),
+        chainId,
+        name: name ?? token?.name ?? '',
+        symbol: symbol ?? token?.symbol ?? '',
+        baseURI: baseURI ?? token?.baseURI ?? '',
+    } as ERC721TokenDetailed
+    cache.set(address, token_)
+    return
```
Jack-Works, line 35: Remove this line

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function useAsyncERC721TokenDetailed(address: string, token?: Partial<ERC721TokenDetailed>) {
+    const chainId = useChainId()
+    const erc721TokenContract = useERC721TokenContract(address) as Contract
+    token = cache.get(address)
+    if (!token) throw ERC721TokenDetailed_(chainId, erc721TokenContract, address, token)
+    return cache.get(address)!
```
Jack-Works, line 43:
```suggestion
    return token
```

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,44 @@
...
+    } as ERC721TokenDetailed
+    cache.set(address, token_)
+    return
+}
+
+export function useAsyncERC721TokenDetailed(address: string, token?: Partial<ERC721TokenDetailed>) {
+    const chainId = useChainId()
+    const erc721TokenContract = useERC721TokenContract(address) as Contract
+    token = cache.get(address)
+    if (!token) throw ERC721TokenDetailed_(chainId, erc721TokenContract, address, token)
```
zhouhanseng, line 42: I think refactor to suspense is not that easy, you need to consider retry and error handle. Please revert it if you are not familiar with suspense.

Jack-Works: Actually, his implementation looks fine currently

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
Jack-Works: Don't mark it as resolved cause it is not resolved

Jack-Works (changes requested):
Still have something we can make better

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -31,14 +22,18 @@ async function ERC721TokenDetailed_(
         symbol: symbol ?? token?.symbol ?? '',
         baseURI: baseURI ?? token?.baseURI ?? '',
     } as ERC721TokenDetailed
-    cache.set(address, token_)
-    return
+
+    if (!cache.has(address)) cache.set(address, token_)
 }
 
-export function useAsyncERC721TokenDetailed(address: string, token?: Partial<ERC721TokenDetailed>) {
```
zhouhanseng: I don't think to integrate with suspense must rename useXXX to getXXX which cause react hooks should be moved out which breaks our code structure, check https://git.io/JtWVq

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,54 @@
...
+export function useAsyncERC721TokenDetailed(
+    address: string,
+    token?: Partial<ERC721TokenDetailed>,
+) {
+    const chainId = useChainId()
+    const erc721TokenContract = useERC721TokenContract(address)
+    const names = ['name', 'symbol', 'baseURI'] as (keyof ERC721['methods'])[]
+    const callDatas = new Array(3).fill([])
+    const [results, calls, _, callback] = useSingleContractMultipleData(erc721TokenContract, names, callDatas)
+    if (!cache.has(address)) throw getERC721TokenDetailed(chainId, address, results, callback, calls, token)
```
guanbinrui, line 48: I haven't written many suspenses before. It is a right way to mix suspense with hook?

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -23,17 +27,28 @@ async function getERC721TokenDetailed_(
...
-    if (!cache.has(address)) throw getERC721TokenDetailed_(chainId, address, results, callback, calls, token)
-    return cache.get(address)!
+    const chainId = useChainId()
+    const erc721TokenContract = useERC721TokenContract(address)
+    const names = ['name', 'symbol', 'baseURI'] as (keyof ERC721['methods'])[]
+    const callDatas = new Array(3).fill([])
+    const [results, calls, _, callback] = useSingleContractMultipleData(erc721TokenContract, names, callDatas)
+    if (!cache.has(address)) throw getERC721TokenDetailed(chainId, address, results, callback, calls, token)
+    try {
+        return cache.get(address)!
```
zhouhanseng, line 50: `cache.get(address)` won't cause any error, `try {} catch {}` is redundant.

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -15,40 +15,30 @@ async function getERC721TokenDetailed(
...
+            type: EthereumTokenType.ERC721,
+            address: formatChecksumAddress(address),
+            chainId,
+            name: name ?? token?.name ?? '',
+            symbol: symbol ?? token?.symbol ?? '',
+            baseURI: baseURI ?? token?.baseURI ?? '',
+        } as ERC721TokenDetailed
+        cache.set(address, token_)
+    } catch {
+        cache.set(address, {} as ERC721TokenDetailed)
```
zhouhanseng, line 32: @lelenei wait, this is not a the best way to cause an error.

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
Jack-Works: that's totally ok. you just don't violate the react hooks rule

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
Jack-Works: Generally, it doesn't matter what it is named. But if you use hooks in it, it must be a hook too.

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
Jack-Works: so why not throw error directly?

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
```diff
@@ -0,0 +1,44 @@
...
+    address: string,
+    results: any[],
+    callback: (calls_: any[]) => Promise<void>,
+    calls: any[],
+    token?: Partial<ERC721TokenDetailed>,
+) {
+    try {
+        await callback(calls)
+        // compose
+        const [name, symbol, baseURI] = results.map((x) => (x.error ? undefined : x.value))
```
Jack-Works, line 21: Hmm, I'm not familiar with this code flow but it seems a little strange to me. You're awaiting `callback(calls)` but don't use the return value of it, instead, you use `results` which comes from `useSingleContractMultipleData` hooks. When code recovered to run, `results` won't be updated (unless that hook is updating the original array directly). So I doubt is it working correctly?

Jack-Works: maybe I should cc @guanbinrui or @zhouhanseng . `results` will be updated in the second pass IIUC but this code is expected to run once for each parameter combination (then it will set the cache, in the second pass, this code pass might not be run cause it already set the cache).

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
zhouhanseng: As I mentioned here: https://git.io/JtWyf

packages/maskbook/src/web3/suspends/getERC721TokenDetailed.ts
zhouhanseng: @lelenei

lelenei: When the useERC721TokenDetailed returns, the content in value is replaced

Jack-Works: let me check out this later

Jack-Works: ok I actually got some knowledge from the react team about how to write suspense code correctly. got change this later

Jack-Works: ok I tried to rewrite this code but found it actually delegate data fetching to `useSingleContractMultipleData`. It therefore should not be an suspense resource. It is useful when you have an async function to get the result. In this case, which is not cause you don't have an async function.

BTW I'm not meaning this is wrong:

```ts
const id = useSomeHook()
return getSuspendedData(id)
```

This is totally correct. But in this case it's not.

```ts
const [results, calls, _, callback] = useSingleContractMultipleData(erc721TokenContract, names, callData)
if (!cache.has(address)) throw getERC721TokenDetailed(chainId, address, results, callback, calls, token)
```

cc @zhouhanseng @guanbinrui @lelenei

packages/maskbook/src/web3/hooks/useERC721TokenDetailed.ts
```diff
@@ -6,19 +6,15 @@ import { useChainId } from './useChainState'
...
     // compose calls
-    const { names, callDatas } = useMemo(
-        () => ({
-            names: ['name', 'symbol', 'baseURI'] as 'name'[],
-            callDatas: new Array(3).fill([]),
-        }),
-        [],
-    )
+    const names = ['name', 'symbol', 'baseURI'] as (keyof ERC721['methods'])[]
+    const callDatas = new Array(3).fill([])
```
Jack-Works, line 17: data cannot have plural.

packages/maskbook/src/plugins/COTM/UI/TokenAlbum.tsx
```diff
@@ -27,26 +28,38 @@ const useStyles = makeStyles((theme) =>
...
+                            </section>
+                        ))}
+                    </div>
+                </div>
+            )}
+        </>
     )
 }
+function COTM_TOKEN_ADDRESS(COTM_TOKEN_ADDRESS: string) {
+    throw new Error('Function not implemented.')
```
Jack-Works, line 64: ?

packages/maskbook/src/plugins/COTM/UI/TokenAlbum.tsx
```diff
@@ -27,26 +28,38 @@ const useStyles = makeStyles((theme) =>
     }),
 )
 
-export interface TokenAlbumProps {}
+export interface TokenAlbumProps {
+    setCollectiblesLoading?: (loading: boolean) => void
```
Jack-Works, line 32: please remove unused props

packages/maskbook/src/plugins/Election2020/UI/ElectionTokenAlbum.tsx
```diff
@@ -28,28 +29,36 @@ const useStyles = makeStyles((theme) =>
     }),
 )
 
-export interface ElectionTokenAlbumProps {}
+export interface ElectionTokenAlbumProps {
+    setCollectiblesLoading?: (loading: boolean) => void
```
Jack-Works, line 33: please remove unused props

Jack-Works (changes requested):

PR #2271, opened 2021-01-21
refactor: integrate usePoolPayload with suspense

zhouhanseng:


packages/maskbook/src/plugins/ITO/UI/ClaimGuide.tsx
```diff
@@ -36,10 +37,13 @@ interface ClaimGuideProps extends Pick<ClaimDialogProps, 'exchangeTokens' | 'pay
 export function ClaimGuide(props: ClaimGuideProps) {
     const { t } = useI18N()
     const { payload, exchangeTokens, isBuyer, open, retryPayload, onClose } = props
+    const [startTransition] = unstable_useTransition({ busyDelayMs: 1000 })
```
Jack-Works, line 40: IIRC `busyDelayMs` is removed in react 👀

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,6 +1,30 @@
-import { useAsyncRetry } from 'react-use'
+import { ValueRef } from '@dimensiondev/holoflows-kit'
 import { PluginITO_RPC } from '../messages'
+import { useValueRef } from '../../../utils/hooks/useValueRef'
+import type { JSON_PayloadInMask } from '../types'
+
+let payloadRef = new ValueRef<JSON_PayloadInMask | null>(null)
```
Jack-Works, line 6: It seems like all `PID`s are shared the same "cache". This is wrong, it will cause a race condition. If your data will never change or outdated, please reference my implementation: `getAssetAsBlobURL`. If it will change, please use https://github.com/vercel/swr (it supports suspense).

packages/maskbook/src/plugins/ITO/UI/ClaimGuide.tsx
zhouhanseng: It is in our codebase.

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
zhouhanseng: I get it! thanks!

zhouhanseng: Thanks, I get it.

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -3,13 +3,13 @@ import { PluginITO_RPC } from '../messages'
 import { useValueRef } from '../../../utils/hooks/useValueRef'
 import type { JSON_PayloadInMask } from '../types'
 
-let payloadRef = new ValueRef<JSON_PayloadInMask | null>(null)
+let storage = new Map<string, ValueRef<JSON_PayloadInMask>>()
```
zhouhanseng, line 6: Here we don't need a cache, retry the pool if needed, so just store each pool by its own pid.

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -4,26 +4,26 @@ import { useValueRef } from '../../../utils/hooks/useValueRef'
...
-    if (isLoading) throw suspender(pid)
+    storageOfLoading.set(pid, true)
+    if (storageOfLoading.get(pid)) throw suspender(pid)
     return
 }
 
 async function suspender(pid: string) {
     try {
         storage.set(pid, new ValueRef(await PluginITO_RPC.getPool(pid)))
-        setTimeout(() => (isLoading = false), 0)
```
zhouhanseng: setTimeout is redundant, removed

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,6 +1,30 @@
-import { useAsyncRetry } from 'react-use'
+import { ValueRef } from '@dimensiondev/holoflows-kit'
 import { PluginITO_RPC } from '../messages'
+import { useValueRef } from '../../../utils/hooks/useValueRef'
+import type { JSON_PayloadInMask } from '../types'
+
+let storage = new Map<string, ValueRef<JSON_PayloadInMask>>()
```
Jack-Works, line 6:
```suggestion
const storage = new Map<string, ValueRef<JSON_PayloadInMask>>()
```

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,6 +1,30 @@
...
-    return useAsyncRetry(() => PluginITO_RPC.getPool(pid), [pid])
+    if (!storageOfLoading.has(pid) || storageOfLoading.get(pid)) throw suspender(pid)
+    return {
+        payload: useValueRef(storage.get(pid)!),
+        retry: () => retry(pid),
+    }
+}
+
+async function retry(pid: string) {
+    storageOfLoading.set(pid, true)
```
Jack-Works, line 18: you can get rid of storageOfLoading, just `storage.delete(pid)` and trigger a new suspended

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,6 +1,28 @@
-import { useAsyncRetry } from 'react-use'
+import { ValueRef } from '@dimensiondev/holoflows-kit'
 import { PluginITO_RPC } from '../messages'
+import { useValueRef } from '../../../utils/hooks/useValueRef'
+import type { JSON_PayloadInMask } from '../types'
+
+const storage = new Map<string, ValueRef<JSON_PayloadInMask>>()
```
Jack-Works, line 6: It seems like you never set on `.value` so you don't need `ValueRef`
```suggestion
const storage = new Map<string, JSON_PayloadInMask>()
```

Jack-Works (approved):

PR #2281, opened 2021-01-24
feat: ITO suspense loading and error retry

zhouhanseng:
Screen record, create an artificial error at first time, click to load successful: 

https://user-images.githubusercontent.com/63733714/105630783-05bf8c00-5e86-11eb-9bd3-0ce8b907f0ab.mov


packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -21,6 +21,6 @@ async function suspender(pid: string) {
     try {
         storage.set(pid, await PluginITO_RPC.getPool(pid))
     } catch (error) {
-        throw error
+        storage.set(pid, {} as JSON_PayloadInMask)
```
zhouhanseng, line 24: This will cause an error in `<ITO />` component which will be captured by its wrapper `<ITO_LoadingFail />`

Jack-Works: what if you throw error directly?

Jack-Works (approved):

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
zhouhanseng: It will cause a dead loop which keep throwing error, and the UI is keeping ITO loading.

PR #2295, opened 2021-01-27
feat: experimental serializer that might have high perf

Jack-Works:
The basic idea is not try to serialize binarys as arrays but into a Blob URL.

The main problem is it might have a bad compatibility because Web Extension is totally unstandardized. Need more investigate in future.

- [x] Works on Chrome
- [?] Works on Firefox
- [?] Works on Firefox for Android
- [?] Works on iOS


Jack-Works:
@septs did you try it?

PR #2302, opened 2021-01-28
chore: more general way for suspense error

zhouhanseng:
go on with https://github.com/DimensionDev/Maskbook/pull/2281

### DIFF
--- a/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
@@ -408,7 +408,7 @@ export function ITO(props: ITO_Props) {
         <div>
             <Card className={classes.root} elevation={0} style={{ backgroundImage: `url(${PoolBackground})` }}>
                 <Box className={classes.header}>
-                    <Typography variant="h5" className={classes.title}>
+                    <Typography variant="h5" className={classes.title} onClick={retryPoolPayload}>
                         {payload.message}
                     </Typography>
 
--- a/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
@@ -184,7 +184,7 @@ export function ITO(props: ITO_Props) {
     const PoolBackground = getAssetAsBlobURL(new URL('../assets/pool-background.jpg', import.meta.url))
 
     const { pid } = props
-    const { payload, retry: retryPoolPayload } = usePoolPayload(pid)
+    const payload = usePoolPayload(pid)
 
     const {
         token,
@@ -304,10 +304,9 @@ export function ITO(props: ITO_Props) {
     //#endregion
 
     const retryITOCard = useCallback(() => {
-        retryPoolPayload()
         retryPoolTradeInfo()
         retryAvailability()
-    }, [retryPoolPayload, retryPoolTradeInfo, retryAvailability])
+    }, [retryPoolTradeInfo, retryAvailability])
 
     const swapStatusText = useMemo(() => {
         if (listOfStatus.includes(ITO_Status.waited)) return t('plugin_ito_status_no_start')
@@ -561,7 +560,7 @@ function ITO_LoadingFailUI({ retryPoolPayload }: { retryPoolPayload: () => void 
     )
 }
 
-export class ITO_LoadingFail extends Component<{ retryPoolPayload: () => Promise<void> }> {
+export class ITO_LoadingFail extends Component<{ retryPoolPayload: () => void }> {
     static getDerivedStateFromError(error: unknown) {
         return { error }
     }
--- a/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
@@ -2,11 +2,10 @@ import { Component, useCallback, useState, useEffect, useMemo } from 'react'
 import classNames from 'classnames'
 import { makeStyles, createStyles, Card, Typography, Box, Link } from '@material-ui/core'
 import { BigNumber } from 'bignumber.js'
-import { isError } from 'lodash-es'
 import { useRemoteControlledDialog } from '../../../utils/hooks/useRemoteControlledDialog'
 import { TransactionStateType } from '../../../web3/hooks/useTransactionState'
 import { WalletMessages } from '../../Wallet/messages'
-import { ITO_Status, JSON_PayloadInMask } from '../types'
+import { ITO_Status } from '../types'
 import { useI18N } from '../../../utils/i18n-next-ui'
 import type { ERC20TokenDetailed, EtherTokenDetailed } from '../../../web3/types'
 import { resolveLinkOnEtherscan } from '../../../web3/pipes'
@@ -186,7 +185,6 @@ export function ITO(props: ITO_Props) {
 
     const { pid } = props
     const { payload, retry: retryPoolPayload } = usePoolPayload(pid)
-    if (isError(payload)) throw new Error()
 
     const {
         token,
--- a/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
@@ -184,7 +184,7 @@ export function ITO(props: ITO_Props) {
     const PoolBackground = getAssetAsBlobURL(new URL('../assets/pool-background.jpg', import.meta.url))
 
     const { pid } = props
-    const payload = usePoolPayload(pid)
+    const { payload, retry: retryPoolPayload } = usePoolPayload(pid)
 
     const {
         token,
@@ -203,7 +203,6 @@ export function ITO(props: ITO_Props) {
     const total = new BigNumber(payload_total)
     const total_remaining = new BigNumber(payload_total_remaining)
     const sold = total.minus(total_remaining)
-
     //#region token detailed
     const {
         value: availability,
@@ -304,9 +303,10 @@ export function ITO(props: ITO_Props) {
     //#endregion
 
     const retryITOCard = useCallback(() => {
+        retryPoolPayload()
         retryPoolTradeInfo()
         retryAvailability()
-    }, [retryPoolTradeInfo, retryAvailability])
+    }, [retryPoolPayload, retryPoolTradeInfo, retryAvailability])
 
     const swapStatusText = useMemo(() => {
         if (listOfStatus.includes(ITO_Status.waited)) return t('plugin_ito_status_no_start')
--- a/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/ITO.tsx
@@ -2,6 +2,7 @@ import { Component, useCallback, useState, useEffect, useMemo } from 'react'
 import classNames from 'classnames'
 import { makeStyles, createStyles, Card, Typography, Box, Link } from '@material-ui/core'
 import { BigNumber } from 'bignumber.js'
+import { isError } from 'lodash-es'
 import { useRemoteControlledDialog } from '../../../utils/hooks/useRemoteControlledDialog'
 import { TransactionStateType } from '../../../web3/hooks/useTransactionState'
 import { WalletMessages } from '../../Wallet/messages'
@@ -27,8 +28,8 @@ import { usePoolTradeInfo } from '../hooks/usePoolTradeInfo'
 import { useDestructCallback } from '../hooks/useDestructCallback'
 import { getAssetAsBlobURL } from '../../../utils/suspends/getAssetAsBlobURL'
 import { EthereumMessages } from '../../Ethereum/messages'
-import { usePoolPayload } from '../hooks/usePoolPayload'
 import { resolveChainName } from '../../../web3/pipes'
+import { usePoolPayload } from '../hooks/usePoolPayload'
 
 export interface IconProps {
     size?: number
@@ -151,8 +152,6 @@ const useStyles = makeStyles((theme) =>
 
 export interface ITO_Props {
     pid: string
-    payload: JSON_PayloadInMask
-    retryPoolPayload: () => Promise<void>
 }
 
 interface TokenItemProps {
@@ -187,6 +186,8 @@ export function ITO(props: ITO_Props) {
 
     const { pid } = props
     const { payload, retry: retryPoolPayload } = usePoolPayload(pid)
+    if (isError(payload)) throw new Error()
+
     const {
         token,
         total: payload_total,
@@ -564,7 +565,6 @@ function ITO_LoadingFailUI({ retryPoolPayload }: { retryPoolPayload: () => void 
 
 export class ITO_LoadingFail extends Component<{ retryPoolPayload: () => Promise<void> }> {
     static getDerivedStateFromError(error: unknown) {
-        console.log('getDerivedStateFromError')
         return { error }
     }
     state: { error: Error | null } = { error: null }
--- a/packages/maskbook/src/plugins/ITO/UI/PostInspector.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/PostInspector.tsx
@@ -6,10 +6,11 @@ export interface PostInspectorProps {
 }
 
 export function PostInspector(props: PostInspectorProps) {
-    const { payload, retry: retryPoolPayload } = usePoolPayload(props.pid)
+    const { retry: retryPoolPayload } = usePoolPayload(props.pid)
+
     return (
         <ITO_LoadingFail retryPoolPayload={retryPoolPayload}>
-            <ITO pid={props.pid} payload={payload} retryPoolPayload={retryPoolPayload} />
+            <ITO pid={props.pid} />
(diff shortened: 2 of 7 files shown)


packages/maskbook/src/plugins/ITO/UI/ITO.tsx
```diff
@@ -187,6 +185,7 @@ export function ITO(props: ITO_Props) {
 
     const { pid } = props
     const { payload, retry: retryPoolPayload } = usePoolPayload(pid)
+    if (payload instanceof Error) throw payload
```
septs, line 188: https://lodash.com/docs/4.17.15#isError

septs (changes requested):

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -21,6 +21,6 @@ async function suspender(pid: string) {
     try {
         storage.set(pid, await PluginITO_RPC.getPool(pid))
     } catch (error) {
-        storage.set(pid, {} as JSON_PayloadInMask)
+        storage.set(pid, new Error())
```
Jack-Works, line 24: a bit strange why don't you just throw an error here?

zhouhanseng: It will cause a dead loop which keep throwing error, and the UI is keeping ITO loading if throw here.

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,7 +1,7 @@
 import { PluginITO_RPC } from '../messages'
 import type { JSON_PayloadInMask } from '../types'
 
-const storage = new Map<string, JSON_PayloadInMask>()
+const storage = new Map<string, JSON_PayloadInMask | Error>()
 
 export function usePoolPayload(pid: string) {
     if (!storage.has(pid)) throw suspender(pid)
```
zhouhanseng, line 7: We also should not throw error in usePoolPayload(), since we call usePoolPayload() outside <ITO /> and <ITO_LoadingFail />

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
Jack-Works: I don't clearly understand why it would happen. If the second render goes normal, it will have no error and goes to line 22 right?

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
Jack-Works: We can add more suspense

packages/maskbook/src/plugins/ITO/UI/ITO.tsx
```diff
@@ -564,7 +566,6 @@ function ITO_LoadingFailUI({ retryPoolPayload }: { retryPoolPayload: () => void
 
 export class ITO_LoadingFail extends Component<{ retryPoolPayload: () => Promise<void> }> {
```
Jack-Works, line 567: So the tree should be like right?
```jsx
<ITO_LoadingFail>
    <Suspense>
```

zhouhanseng: Tree is like this since `usePoolPayload()` need to be inside `<Suspense>`
```
<Suspense>
  <ITO_LoadingFail>
    <ITO />
```

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
zhouhanseng: Do not need more suspense.

```xml
<Suspense>
  <PostInspector> // usePoolPayload() 
    <ITO_LoadingFail>
      <ITO>
```

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
zhouhanseng: Yes, if `await PluginITO_RPC.getPool(pid)` works well.

zhouhanseng: you mean that screen video, I make the error intentionally 

```ts
i = 0

async function suspender(pid: string) {
  try {
    if (i === 0) {
      i++
      throw new Error()
    }
    storage.set(pid, await PluginITO_RPC.getPool(pid))
  } catch {
    storage.set(pid, new Error())
  }
}
```

Jack-Works: Ok, this makes sense I understand why now. https://github.com/facebook/react/issues/17526

Jack-Works: So according to the link above, to make things correct, we should do like this:

```js
const cache = storage.get(...)
// cache match
if (isError(cache)) throw cache // Must sync
else return cache

throw getPool(pid)
```

But I got no idea how to clear the error cache when recovering from the ErrorBoundary

Jack-Works: Maybe like this: Only throw once then clear the error one? Or clear the whole cache?

```js
if (isError(cache)) { storage.remove(pid); throw cache }
```

Jack-Works: The problem here is that the storage is global so it won't be cleared out in the error recovery case

Jack-Works: Ok I think this can resolve the problem:

```js
const storage = new Map()
function useSomething(id) {
    const [fatalError, setError] = useState(null)
    const result = storage.get(id)
    if (result) return result
    if (fatalError) throw fatalError
    if (!result) throw suspender(id).catch(setError)
}
```

Storage error in this way, we can automatically clear the error when the component tree is recovered from the ErrorBoundary. @zhouhanseng would you like to have a try?

zhouhanseng:
```ts
if (isError(cache)) { storage.remove(pid); throw cache }
```

I don't quite follow from those code snippets.

zhouhanseng: @Jack-Works sure

zhouhanseng: @Jack-Works The code below doesn't work as expection, 
Expect: it would display the <ITO_Failue />
Actual: always show <ITO_Loading />

```ts
import { PluginITO_RPC } from '../messages'
import type { JSON_PayloadInMask } from '../types'
import { useState } from 'react'

const storage = new Map<string, JSON_PayloadInMask | Error>()

export function usePoolPayload(pid: string) {
    const [fatalError, setError] = useState(null)
    const result = storage.get(pid)

    if (fatalError) throw fatalError

    if (!result) {
        throw suspender(pid).catch(setError)
    }

    return {
        payload: storage.get(pid)!,
        retry: () => retry(pid),
    }
}

async function retry(pid: string) {
    storage.delete(pid)
    if (!storage.has(pid)) throw suspender(pid)
    return
}

async function suspender(pid: string) {
    try {
        // if (i === 100) {
            // i++
        throw new Error()
        // }
        // storage.set(pid, await PluginITO_RPC.getPool(pid))
    } catch (error) {
        throw error
        // storage.set(pid, new Error())
    }
}

```

Jack-Works:
```ts
import { PluginITO_RPC } from '../messages'
import type { JSON_PayloadInMask } from '../types'
import { useState } from 'react'

const storage = new Map<string, JSON_PayloadInMask>()

export function usePoolPayload(pid: string) {
    const [fatalError, setError] = useState(null)
    const result = storage.get(pid)

    if (result) return {
        payload: storage.get(pid)!,
        retry: () => retry(pid),
    }
    if (fatalError) throw fatalError
    throw suspender(pid).catch(setError)
}

async function retry(pid: string) {
    storage.delete(pid)
    throw suspender(pid)
}

async function suspender(pid: string) {
    return storage.set(pid, await PluginITO_RPC.getPool(pid))
}
```

Try this one @zhouhanseng

zhouhanseng: @Jack-Works 

```ts
async function suspender(pid: string) {
    throw new Error()
}
```

Still always show <ITO_Loading />

Jack-Works: @zhouhanseng please tell me where can I invoke this hook, I will look at this later today

zhouhanseng: https://twitter.com/Zhouhancheng/status/1352487007411269632

Jack-Works:
@zhouhanseng I have pushed one commit. It works well on my side. plz try it

zhouhanseng:
@Jack-Works Hi,

```
async function suspender(pid: string) {
    throw new Error()
}
```
This still keeps loading.

```
const [e, set] = useState<Error | null>(new Error())
```

And this will cause

Jack-Works:
If you always throw an error in the suspend it of course will load infinitely. Please try this way: `if (Math.random() > 0.5) throw new Error('simulate error')`

zhouhanseng:
I see, then there's no error status at all, it will retry auto when error.

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,26 +1,22 @@
 import { PluginITO_RPC } from '../messages'
 import type { JSON_PayloadInMask } from '../types'
 
-const storage = new Map<string, JSON_PayloadInMask>()
-
+const cache = new Map<string, [0, Promise<void>] | [1, JSON_PayloadInMask] | [2, Error]>()
+export function poolPayloadRetry() {
+    cache.forEach(([status], pid) => status === 2 && cache.delete(pid))
+}
```
zhouhanseng, line 7: Precisely this is `poolPayloadErrorRetry()`

packages/maskbook/src/plugins/ITO/UI/ITO.tsx
```diff
@@ -305,10 +304,9 @@ export function ITO(props: ITO_Props) {
     //#endregion
 
     const retryITOCard = useCallback(() => {
-        retryPoolPayload()
         retryPoolTradeInfo()
         retryAvailability()
-    }, [retryPoolPayload, retryPoolTradeInfo, retryAvailability])
```
zhouhanseng, line 311: We need a `retryPoolPayload()` when there's no error.

Jack-Works: You can add a separate hooks that removes the cache and trigger an update. I'll make this pattern mote general later

zhouhanseng: I'd like to add a `forceUpdate()`

https://stackoverflow.com/a/55862077/4803321

nuanyang233: Is it best practice to use forceUpdate here? I don't know enough about Suspense, but forceUpdate gives me a gut feeling that it will cause the whole component tree to re-render🤔

packages/maskbook/src/utils/hooks/useForceUpdate.ts
```diff
@@ -0,0 +1,9 @@
+import { useState, useCallback } from 'react'
+
+export function useForceUpdate() {
```
Jack-Works, line 3: Please use `useUpdate` from `react-use` instead of inventing one

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,26 +1,32 @@
...
-    return
-}
-
 async function suspender(pid: string) {
-    try {
-        storage.set(pid, await PluginITO_RPC.getPool(pid))
-    } catch (error) {
-        storage.set(pid, {} as JSON_PayloadInMask)
-    }
+    if (Math.random() > 0.1) throw new Error()
```
Jack-Works, line 30: hmm remove this line?

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,26 +1,32 @@
...
+export function poolPayloadErrorRetry() {
+    cache.forEach(([status], pid) => status === 2 && cache.delete(pid))
+}
 export function usePoolPayload(pid: string) {
-    if (!storage.has(pid)) throw suspender(pid)
-    return {
-        payload: storage.get(pid)!,
-        retry: () => retry(pid),
+    const rec = cache.get(pid)
+    const forceUpdate = useForceUpdate()
```
Jack-Works, line 11: use `useUpdate` from `react-use`

packages/maskbook/src/plugins/ITO/UI/PostInspector.tsx
```diff
@@ -1,15 +1,15 @@
...
-import { usePoolPayload } from '../hooks/usePoolPayload'
 
 export interface PostInspectorProps {
     pid: string
 }
-
+const { log } = console
 export function PostInspector(props: PostInspectorProps) {
-    const { payload, retry: retryPoolPayload } = usePoolPayload(props.pid)
+    log('post inspector')
```
Jack-Works, line 9: remove the log?

packages/maskbook/src/plugins/ITO/UI/ITO.tsx
nuanyang233: Please ignore it, I've thought about it for a while and there seems to be no better way to trigger update here

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,26 +1,32 @@
...
+            .catch((e) => void cache.set(pid, [2, e]))
+        cache.set(pid, [0, p])
+        throw p
     }
+    if (rec[0] === 1)
+        return {
+            payload: rec[1],
+            retry: () => {
+                if (cache.has(pid)) cache.delete(pid)
+                forceUpdate()
```
Jack-Works, line 24: Please wrap this call in the `startTransition` (see docs at https://reactjs.org/docs/concurrent-mode-reference.html#usetransition).

Jack-Works (changes requested):
Generally LGTM, some minor changes requested

packages/maskbook/src/plugins/ITO/UI/ITO.tsx
Jack-Works: @nuanyang233 yes it will let the whole tree re-render and trigger new suspense. If we wrap it in `startTransition`, it will make the process smoother.

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
zhouhanseng: I'm not sure if we should coupling startTransition with retry. In some cases, maybe we want to see the loading state.

We can use startTransition on where we want to use.

https://github.com/DimensionDev/Maskbook/blob/master/packages/maskbook/src/plugins/ITO/UI/ClaimGuide.tsx#L42

Jack-Works: Fine, if you think it is a necessity to show the loading state then this LGTM

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
```diff
@@ -1,26 +1,34 @@
...
+            .then((val) => void cache.set(pid, [1, val]))
+            .catch((e) => void cache.set(pid, [2, e]))
+        cache.set(pid, [0, p])
+        throw p
     }
+    if (rec[0] === 1)
+        return {
+            payload: rec[1],
+            retry: () =>
+                startTransition(() => {
```
Jack-Works, line 25: If you think we should show the loading state directly you can remove the `startTransition`. BTW `useTransition` returns a `isPending` argument so you can return it as `isRefeshing` and still show the old data with a loading refresh hint

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
zhouhanseng: okay

packages/maskbook/src/plugins/ITO/hooks/usePoolPayload.ts
Jack-Works: Hmm wait, I think this code is not ready for `startTransition` cause it is using the global state (rely on `cache`) so remove it if you like. I'll provide an infra later that fully works with these whole bunch of things

Jack-Works (approved):

Jack-Works (approved):

PR #2304, opened 2021-01-28
chore: update redpacket contract

zhouhanseng:


packages/maskbook/src/plugins/RedPacket/types.ts
```diff
@@ -33,7 +33,7 @@ export interface RedPacketAvailability {
     total: string
     claimed: string
     expired: boolean
-    ifclaimed: boolean
+    claimed_amounts: number
```
Jack-Works, line 36: Will it break the current code? Does this object published as a payload or stored in the database?

packages/maskbook/src/plugins/RedPacket/constants.ts
```diff
@@ -7,8 +7,8 @@ export const RedPacketPluginID = 'com.maskbook.red_packet'
 export const RED_PACKET_CONSTANTS = {
     HAPPY_RED_PACKET_ADDRESS: {
         [ChainId.Mainnet]: '0x26760783c12181efa3c435aee4ae686c53bdddbb',
```
Jack-Works, line 9: mainnet doesnt update, does it mean mainnet still using the old contract?

packages/maskbook/src/plugins/RedPacket/types.ts
zhouhanseng: Neither of them, actually `ifclaimed` has not been used before. And this object will be calculated every time when load the red packet to decide the status.

packages/maskbook/src/plugins/RedPacket/constants.ts
zhouhanseng: ye

guanbinrui (approved):

yisiliu:
I don't get it - why would you @zhouhanseng merge this into the master without deploying the new contract?

zhouhanseng:
Lack of experience. This PR should not merge into master, otherwise it's possible to be released.

PR #2325, opened 2021-02-01
docs: give general guides on the use of css-in-js

Jack-Works:
### General guides

-   ✅ For recommendations
-   ⚠ For warnings
-   🚫 For forbiddens
-   ✅ Use [the Box component provided by the library](https://next.material-ui.com/components/box/#main-content) when the CSS is simple and only used once.
-   ✅ CSS custom variables is OK but do not abuse it. Get the variable from the theme if it is possible.
-   ✅ CSS grid is OK but you may want to use the `Grid` component. Choose the fit one based on your need.
-   🚫 DO NOT use mystery abbreviations in the `Box` component, e.g. `<Box sx={{ p: 5 }} />` (DONT) but `<Box sx={{ padding: 5 }}>` (DO, easier to read).
-   🚫 DO NOT mix two styling solution together. It may cause unknown bugs. FYI: `makeStyles` is `JSS`, `Box` and `styled` are `@emotion`.

### Guides on the `makeStyles` style

```js
conts useStyle = makeStyles(theme => ({
    root: { margin: theme.spacing(4) }
}))
```

#### ✅⚠ You can use `makeStyles` in the project

Reason: This is the recommend way of writing way in @material-ui 4.

⚠ In @material-ui 5 the recommend way has changed to the `styled-component` style but that style is not easy to style the inner element.

⚠ I (@Jack-Works) will try to convince the library team to make a emotion styled version makeStyles or make it on my own so we can get rid of 2 different CSS-in-JS framework in our project.

### Guides on the `styled` style (let's call it `styled component`)

```js
const Title = style.div`
    display: flex;
`;
```

#### ✅ You can use `styled component` in the project.

Reason: Good DX

#### ⚠ You may want to use "object style" when it involves dynamic styles

Reason: _Potential_ performance lost.

Not enforcement because it is not confirmed.

```js
// ⚠ Not so good
const Title = style(Typography)(
    ({ theme }) => `
    marign-left: ${theme.spacing(4)};
`
);

// ✅ OK
const Title = style(Typography)(({ theme }) => {
    marginLeft: theme.spacing(4);
});
```

#### 🚫 Direct selector to the deeper elements of the library

```js
const Title = style.div`
    & .Mui-selected { display: flex; }
`;
```

DO NOT do this.

Reason: Not type-checked. Easy to get things wrong.

#### 🚫⚠ Calculated selector to the deeper elements of the library

```js
import { listItemClasses } from "@material-ui/core";
const Title = style.div`
    & .${listItemClasses.selected} {
        display: flex;
    }
`;
```

🚫 DO NOT use this pattern in the content script.

⚠ Not suggested to use in other UI context.

Reason: Though it is type checked, it actually not safe in our content script environment. In the content script we make the styles generated by the emotion randomized. This method is actually based on the global name in the library and not guaranteed to work in our case.

#### ✅ `components` or `*Component` style

Overwriting styles in this way is good.

```js
const MyExample = styled(Example)`
    /* CSS goes here */
`

// Use it via:
<Parent components={{ example: MyExample }} />
// Or:
<Parent ExampleComponent={MyExample} />
```

⚠ No every component has this kind of API that allowing to overwrite the inner component so this method may not be able to use.


septs:
1. move caveats.md to docs folder
2. update readme file

PR #2360, opened 2021-02-05
fix: token amount convert precision

zhouhanseng:
fix slider undraggable bug

fix the bug set tokenAmount by click max button not equal to slide to the max.

### DIFF
--- a/packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
@@ -184,8 +184,7 @@ export function ClaimDialog(props: ClaimDialogProps) {
         if (claimAmount.isEqualTo(0)) return t('plugin_ito_error_enter_amount')
         if (claimAmount.isGreaterThan(new BigNumber(tokenBalance)))
             return t('plugin_ito_error_balance', { symbol: claimToken.symbol })
-        if (claimAmount.dividedBy(ratio).isGreaterThan(maxSwapAmount))
-            return t('plugin_ito_dialog_claim_swap_exceed_wallet_limit')
+        if (tokenAmount.isGreaterThan(maxSwapAmount)) return t('plugin_ito_dialog_claim_swap_exceed_wallet_limit')
         return ''
     }, [claimAmount, tokenBalance, maxSwapAmount, claimToken, ratio])
 
--- a/packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
@@ -133,6 +133,14 @@ export function ClaimDialog(props: ClaimDialogProps) {
     )
     //#endregion
 
+    //#region maxAmount for TokenAmountPanel
+    const maxAmount = useMemo(() => BigNumber.min(maxSwapAmount.multipliedBy(ratio).dp(0), tokenBalance).toFixed(), [
+        maxSwapAmount,
+        ratio,
+        tokenBalance,
+    ])
+    //#endregion
+
     //#region claim
     const [claimState, claimCallback, resetClaimCallback] = useClaimCallback(
         payload.pid,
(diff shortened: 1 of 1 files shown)


packages/maskbook/src/plugins/Wallet/formatter.ts
```diff
@@ -26,8 +26,9 @@ export function formatBalance(balance: BigNumber, decimals: number, significant:
 
     let fraction = balance.modulo(base).toString(10) // (balance % base).toString(10)
 
+    const fractionFloatLength = fraction.match(/(\.\d+)?$/)?.[0].length ?? 0
```
zhouhanseng, line 29: You may think there should not be balance with float number, the minimum unit of a token is always 1. But actually the balance we pass into `formatBalance()` is not always directly from a token balance, but a calculating result of bigNumber. E.g calculate swap token by the ratio.

Jack-Works: so why not add it to the code as comment 👀

septs: https://mikemcl.github.io/bignumber.js/#toFr

zhouhanseng: `toFraction()` is not suitable in our case.

guanbinrui: The `formatBalance` should only handle integers.

zhouhanseng: OK, convert all floats to integers before pass to formatBalance.

packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
```diff
@@ -224,15 +231,11 @@ export function ClaimDialog(props: ClaimDialogProps) {
                     className={classes.swapLimitSlider}
                     value={Number(tokenAmount.dividedBy(maxSwapAmount).multipliedBy(100))}
                     onChange={(_, newValue) => {
-                        let tAmount = maxSwapAmount.multipliedBy((newValue as number) / 100)
-                        const swapAmount = formatBalance(tAmount.multipliedBy(ratio), claimToken.decimals)
```
zhouhanseng, line 228: After adopt `BigNumber.dp(0)` in https://github.com/DimensionDev/Maskbook/pull/2365, remove the incorrect `formatBalance()` which should not be used in the process of token conversion calculating. Then the slider undraggable bug fixed

packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
```diff
@@ -206,8 +214,7 @@ export function ClaimDialog(props: ClaimDialogProps) {
         if (claimAmount.isEqualTo(0)) return t('plugin_ito_error_enter_amount')
         if (claimAmount.isGreaterThan(new BigNumber(tokenBalance)))
             return t('plugin_ito_error_balance', { symbol: claimToken.symbol })
-        if (claimAmount.dividedBy(ratio).isGreaterThan(maxSwapAmount))
```
zhouhanseng, line 209: Compare token amount state with state directly rather than with the calculating result which is not precise.

E.g. after click the max button, or input the max number or slide to the max




```ts
console.log(tokenAmount.toFixed()) // 601424451774961270 ✅
console.log(maxSwapAmount.toFixed()) // 601424451774961270 ✅
console.log(claimAmount) // 79388028 ✅
console.log(claimAmount.dividedBy(ratio).toFixed()) // 601424454545454545.45454545454545454545 ❌
``` 

So, why `claimAmount.dividedBy(ratio) !== tokenAmount` ? 
[ClaimDialog.tsx/R137](https://github.com/DimensionDev/Maskbook/pull/2360/files#diff-30e825f79dbac68533e6dfe3b6c19c4baf2fba760b3631a5f83d9404d0da491eR137)

```ts
const maxAmount = useMemo(() => BigNumber.min(maxSwapAmount.multipliedBy(ratio).dp(0), tokenBalance).toFixed(), 
```

We lose the precision by call `.dp(0)`, so why call `dp(0)` here?

[TokenAmountPanel.tsx#L145](https://github.com/DimensionDev/Maskbook/blob/master/packages/maskbook/src/web3/UI/TokenAmountPanel.tsx#L145): 

```
formatBalance(new BigNumber(maxAmount ?? balance), token.decimals)
```

Because we need to pass maxAmount to formatBalance which only handles integer.

I think we may should refactor props of `<TokenAmountPanel />` from, because set a value which indicates BigNumber with string type by using `formatBalance()` or `.toFixed()` could lead to precision losing.

```ts
amount: string
maxAmount?: string
balance: string
```

to: 

```ts
amount: BigNumber
maxAmount?: BigNumber
balance: BigNumber
```

packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
```diff
@@ -256,8 +259,11 @@ export function ClaimDialog(props: ClaimDialogProps) {
                         value === ''
                             ? new BigNumber(0)
                             : new BigNumber(value).multipliedBy(new BigNumber(10).pow(claimToken.decimals))
-                    setClaimAmount(val)
-                    setTokenAmount(val.dividedBy(ratio))
+                    const isMax = value === formatBalance(new BigNumber(maxAmount), claimToken.decimals)
```
zhouhanseng, line 262: Fix with current `<TokenAmountPanel />`, we will lose some precision which lead to 

A. click max button or input max number



B. using slider

zhouhanseng:
Test here ropsten https://twitter.com/Zhouhancheng/status/1358257669228752896

packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
guanbinrui: I noticed that `dp(0)` will use the configured `ROUNDING_MODE` by default. And in this case, `79388027.63429488764` seems rounded to `79388028` which becomes a larger number. I don't know if this is a conversion way for a finance-related app? @xiazhiming Any advice?

zhouhanseng: Notice that MSKD is 6 decimals, so 79.388028 (79388028) is right, there's no `79388027.63429488764` for MSKD. So if there's no `isMax` handling, you can't get all 0.60142445177496127 MSKA from `<TokenAmountPanel />`, the largest MSKA you can get is to input 79.388027. 79.388028 * raito would greater than 0.60142445177496127

guanbinrui (approved):

PR #2365, opened 2021-02-05
fix: rm swap amount decimal when slide

zhouhanseng:
Remove the unacceptable decimal for the calculating result of swap token when slide.

the unacceptable decimal would cause error when approve token:

```
console.log(tAmount.multipliedBy(ratio)) 
console.log(tAmount.multipliedBy(ratio).dp(0)) 

// 79388027.0000000000000000000000004
// 79388027
```

### DIFF
--- a/packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
+++ b/packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
@@ -230,8 +230,8 @@ export function ClaimDialog(props: ClaimDialogProps) {
                             .dividedBy(ratio)
                             .multipliedBy(Math.pow(10, claimToken.decimals))
                         if (tAmount.isGreaterThan(maxSwapAmount)) return
-                        setTokenAmount(tAmount)
-                        setClaimAmount(tAmount.multipliedBy(ratio))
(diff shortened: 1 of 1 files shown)


packages/maskbook/src/plugins/ITO/UI/ClaimDialog.tsx
```diff
@@ -230,8 +230,8 @@ export function ClaimDialog(props: ClaimDialogProps) {
                             .dividedBy(ratio)
                             .multipliedBy(Math.pow(10, claimToken.decimals))
                         if (tAmount.isGreaterThan(maxSwapAmount)) return
-                        setTokenAmount(tAmount)
-                        setClaimAmount(tAmount.multipliedBy(ratio))
+                        setTokenAmount(tAmount.dp(0))
+                        setClaimAmount(tAmount.multipliedBy(ratio).dp(0))
```
zhouhanseng, line 234: Only call `dp(0)` when set the state rather than on each BigNumber multiply or divide calculation, otherwise it would lose precision. If call `dp(0)` on every calculation when slide, you will find the slide undraggable bug solved https://github.com/DimensionDev/Maskbook/pull/2360, but the calculating result loses precision.

guanbinrui (approved):

PR #2404, opened 2021-02-15
fix: mm locked state workaround

zhouhanseng:
Issue: https://github.com/MetaMask/metamask-extension/issues/10368

`inpage-provider` pkg workaround: https://github.com/DimensionDev/extension-provider/pull/5

Fix the issue that when metamask has locked, but ITO still shown, which leads to errors in the further using:

Now with https://github.com/DimensionDev/extension-provider/pull/5, we can detect the lock event. unlock event still can't be detected, but it's almost okay since ITO displays as below image as soon as lock happens(whether manually or automatically lock), click the button to connect the metamask and the following unlock event can be detected, because `Services.Ethereum.connectMetaMask()` creates a new provider first:

### DIFF
--- a/packages/maskbook/src/extension/background-script/EthereumServices/providers/MetaMask.ts
+++ b/packages/maskbook/src/extension/background-script/EthereumServices/providers/MetaMask.ts
@@ -21,9 +21,10 @@ async function onAccountsChanged(accounts: string[]) {
     currentIsMetamaskLockedSettings.value = !(await provider!._metamask?.isUnlocked()) && accounts.length === 0
 }
 
-function onChainIdChanged(id: string) {
+async function onChainIdChanged(id: string) {
     // learn more: https://docs.metamask.io/guide/ethereum-provider.html#chain-ids and https://chainid.network/
     const chainId = Number.parseInt(id, 16) as ChainId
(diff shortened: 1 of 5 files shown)


packages/maskbook/src/extension/background-script/EthereumServices/providers/MetaMask.ts
```diff
@@ -21,9 +21,10 @@ async function onAccountsChanged(accounts: string[]) {
     currentIsMetamaskLockedSettings.value = !(await provider!._metamask?.isUnlocked()) && accounts.length === 0
 }
 
-function onChainIdChanged(id: string) {
+async function onChainIdChanged(id: string) {
     // learn more: https://docs.metamask.io/guide/ethereum-provider.html#chain-ids and https://chainid.network/
     const chainId = Number.parseInt(id, 16) as ChainId
+    currentIsMetamaskLockedSettings.value = !(await provider!._metamask?.isUnlocked())
```
zhouhanseng, line 27: After locked, then if unlock the metamask from metamask app directly, then click the "connect metamask wallet" button would not trigger `onAccountsChanged`, but `onChainIdChanged` would still be triggered. So set `currentIsMetamaskLockedSettings` here.

guanbinrui (approved):

PR #2413, opened 2021-02-18
Create zh-hant chinese language

ystartgo:
add zh-hant chinese language messages.json


septs:
plz update `packages/maskbook/src/_locales/zh/messages.json`.

Jack-Works:
yeah, our default zh is zh-hant so please update that file

PR #2468, opened 2021-02-23
fix: ci build

septs:


guanbinrui (approved):

package.json
```diff
@@ -11,7 +11,7 @@
     "start:storybook": "dev -- start-storybook -p 9009 -s public --quiet",
     "build": "build",
     "build-e2e": "env target=E2E build",
-    "build-ci": "node --max_old_space_size=4096 -r ts-node/register/transpile-only scripts/ci-build.ts",
+    "build-ci": "ts-node scripts/ci-build.ts",
```
Jack-Works, line 14: so the problem of OOM no longer appears now?

septs: https://nodejs.org/en/knowledge/child-processes/how-to-spawn-a-child-process/

PR #2475, opened 2021-02-24
feat: add css vars for mask colors

Jack-Works:


nuanyang233:
hmmm, I rebase this branch on the new-wallet branch and do some tested. If we use this approach we don't seem to be able to use the colors of the light theme in the dark theme, like this.

image

I understand is that we will set `textNormal` in `LightColor` and `DarkColor`, and then take different colors according to the color mode.

Jack-Works:
Yes, you should not use a light theme color in a dark theme. This is a feature

nuanyang233:
@Jack-Works cc

PR #2504, opened 2021-02-26
fix and add Japanese locale

blank71:
- `暗号通貨ウォレット` is too long so I changed it to `ウォレット`.
- Line 381, 382 is not translated. Will you explain me `Route` and `Portion`?
- Line 476: I translated `Infinity` to `無限(no limited)`. Is that Okay? `Infinity` can be translated to `永遠(forever)` so.
- `Persona` is now translated to `人格`. But this word is not familiar to Japanese people. Instead, `ペルソナ` is better than `人格`. Although also `ペルソナ` is not familiar to Japanese people too.


Tedko (approved):
The game Persona is cool.

Tedko:
無限(no limited) is better! @blank71

PR #2512, opened 2021-03-01
fix: create ito dialog tab

zhouhanseng:
1. click send in history.
2. enter ito dialog


packages/maskbook/src/plugins/ITO/UI/CompositionDialog.tsx
```diff
@@ -32,6 +32,9 @@ export function CompositionDialog(props: CompositionDialogProps) {
     const account = useAccount()
     const chainId = useChainId()
 
+    //#region tabs
+    const state = useState(0)
```
Jack-Works, line 36: Maybe you should limit the value domain by `useState<0 | 1 | 2>(0)` etc. give them a meaningful name is better

Jack-Works (approved):

PR #2513, opened 2021-03-01
chore: ui optimise toolbar shrink

zhouhanseng:
Expand or shrink tool bar should not re-fetch token.

### DIFF
--- a/packages/maskbook/src/components/InjectedComponents/ToolbarAtTwitter.tsx
+++ b/packages/maskbook/src/components/InjectedComponents/ToolbarAtTwitter.tsx
@@ -124,45 +124,44 @@ export function ToolbarAtTwitter(props: ToolbarAtTwitterProps) {
     }, [windowWidth])
     //#endregion
 
+    const IconButtonMemo = useMemo(
+        () => () => (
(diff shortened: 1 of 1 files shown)


packages/maskbook/src/components/InjectedComponents/ToolbarAtTwitter.tsx
```diff
@@ -124,45 +124,44 @@ export function ToolbarAtTwitter(props: ToolbarAtTwitterProps) {
     }, [windowWidth])
     //#endregion
 
+    const IconButtonMemo = useMemo(
```
zhouhanseng, line 127: There's an animation effect when click button which would keep showing after expand or shrink, so re-render IconButton immediately when isExpand changes.

guanbinrui (approved):

packages/maskbook/src/components/InjectedComponents/ToolbarAtTwitter.tsx
Jack-Works: so why not leave it in the code?🤔

PR #2519, opened 2021-03-02
chore: lock package version

zhouhanseng:


package.json
```diff
@@ -71,7 +71,7 @@
     "ts-node": "^9.1.1",
     "ttypescript": "^1.5.12",
     "type-coverage": "^2.14.8",
-    "typescript": "^4.1.3"
+    "typescript": "4.1.3"
```
zhouhanseng, line 74: typescript@4.2.2 has bug

PR #2530, opened 2021-03-04
feat: new api for cross-env settings

### LINKED ISSUE #1884: [Demand] Settings/state internal implementation need an isolated API
Now we're using `createGlobalSettings` for the internal state (for example, currently connected wallet type) which is a misuse. Because `createGlobalSettings` will persistent the stored value but it may not make sense for things like `currently connected wallet type`.

|                          | Persistent ✔         | Persistent ❌      |
|--------------------------|----------------------|-------------------|
| Need sync across context | createGlobalSettings | \<New API we need> |
| No need to sync          | N/A                  | ValueRef          |

Jack-Works:
close #1884 

This is a draft. You can see the API shape (RFC cc @guanbinrui @septs @zhouhanseng @nuanyang233) on the usage site at packages/shared/src/state/use-site.ts

The new API is generally compatible with the current API shape but

- React concurrent-mode safe (version field, useMutableSource)
- Allow split the API model so isolate mode dashboard can use those infras
- Add non-persistent storage (see #1884)

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@dimensiondev/maskbook-shared",
   "dependencies": {
+    "@dimensiondev/holoflows-kit": "0.8.0-20201215084554-2f9e757",
     "async-call-rpc": "^5.0.0",
     "bignumber.js": "^9.0.1",
     "immer": "^8.0.1",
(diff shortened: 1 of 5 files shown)


Jack-Works:
gonna merge this since there is no use site of this and it won't break anything

PR #2545, opened 2021-03-09
chore: add icons view box prop

nuanyang233:


packages/icons/general/Card.tsx
```diff
@@ -11,4 +11,5 @@ export const CardIcon = createIcon(
             </clipPath>
         </defs>
     </>,
+    '0 0 16 16',
```
zhouhanseng, line 14: Should we add a fixed viewbox? What about the icon width or height is greater than 16?

nuanyang233: yes, we need set the correct viewbox as in the design. we can set the icon container font-size and set `fontSize='inherit` or the default size (eg: large, small)

zhouhanseng (approved):

Jack-Works (approved):

PR #2581, opened 2021-03-14
Issue: Enabling Partial Encryption

mountainfirefly:


packages/maskbook/src/components/InjectedComponents/PostDialog.tsx
```diff
@@ -155,18 +156,28 @@ export function PostDialogUI(props: PostDialogUIProps) {
...
                             }}
+                            // ref={(el) => inputRef.current = el}
                             autoFocus
                             value={props.postContent.content}
                             onChange={onPostContentChange}
                             fullWidth
                             multiline
                             placeholder={t('post_dialog__placeholder')}
                             inputProps={{ 'data-testid': 'text_textarea' }}
+                            inputRef={(el) => inputRef.current = el}
```
septs, line 167:
```suggestion
                            inputRef={inputRef}
```

packages/maskbook/src/social-network-provider/facebook.com/worker-provider.ts
```diff
@@ -1,7 +1,7 @@
 import { defineSocialNetworkWorker } from '../../social-network/worker'
 import { sharedProvider } from './shared-provider'
-import { fetchPostContentFacebook } from './worker/fetchPostContent'
-import { fetchProfileFacebook } from './worker/fetchProfile'
+import { fetchPostContentFacebook } from './Worker/fetchPostContent'
+import { fetchProfileFacebook } from './Worker/fetchProfile'
```
septs, line 4: CC @Jack-Works 

plz use `git mv` command, rename folder name
see https://git-scm.com/docs/git-mv

packages/maskbook/src/components/InjectedComponents/PostDialog.tsx
mountainfirefly: Sure

packages/maskbook/src/social-network-provider/facebook.com/worker-provider.ts
mountainfirefly: Sure

mountainfirefly: Done, please check.

septs:
plz rebase master

Jack-Works:
Hi! @mountainfirefly thanks for your contribution. For this specific problem, you may want to wait for our SocialNetwork refactoring (in progress!) which involves a great number of changes in the codebase.

And you may want to know how to do this in a correct design (TypedMessage related).

mountainfirefly:
Hey @Jack-Works,

Thank you and sure I can give it a try if you can provide me the design guidelines.

Jack-Works:
@mountainfirefly hi sorry for the delay. so here is the guideline #2581

mountainfirefly:
@Jack-Works You have attached the current issue link mistakenly I think, it is not the guideline. Can you please check?

Jack-Works:
oh sorry, that is the "guideline" or the "blueprint" but it seems pretty abstract to the community developers. let me explain, here are two ways do to partial encryption.

The first way is a quick way, you do the minimal changes to the composition UI and other parts remain unchanged. The second way is the correct way (that issue points to) once the inrfa I mentioned in that issue completed we will naturally have the functionality of partial encryption and other more powerful features around the composition.

mountainfirefly:
Hey, I didn't get what you are trying to say. 
Isn't it the right way how I have done it in this PR and here is a [demo](https://www.youtube-nocookie.com/embed/krHj4zaXhNk)?

yisiliu:
> Hey @Jack-Works, I didn't get what you are trying to say.
> Isn't it the right way how I have done it in this PR and here is a [demo](https://www.youtube-nocookie.com/embed/krHj4zaXhNk)?

Hey @mountainfirefly this demo seems pretty interesting but I would like to address a few improvements, temporarily ignoring the development guidelines introduced by @Jack-Works :
1. I think it's better if you can add a button somewhere so that anyone with the proper access can click to revel the encrypted text
2. It should also allow multi-spot partial encryptions so that you can really realize a cloze test on Twitter
3. Based on the above, I think it's better to add the button on the blanked out area

What do you think?

Jack-Works:
> Hey @Jack-Works, I didn't get what you are trying to say.
> Isn't it the right way how I have done it in this PR and here is a [demo](https://www.youtube-nocookie.com/embed/krHj4zaXhNk)?

Interesting demo, but I think we should reverse the button, which means, the default text is encrypted and if you want to make it unencrypted, you select the text and use that button. how do u think @yisiliu

mountainfirefly:
Sure @Jack-Works @yisiliu, I'll think about it but it requires changes on the backend as well.

Jack-Works:
> I'll think about it but it requires changes on the backend as well.

what do you mean by the "backend"?

PR #2584, opened 2021-03-15
chore: improve locale-kit

septs:


guanbinrui (approved):

Jack-Works (approved):

packages/locale-kit/locale-kit.ts
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env ts-node
```
Jack-Works, line 1: Does it work on Windows?

septs: Migrate `cli-dev-build` to TypeScript?

PR #2595, opened 2021-03-15
chore: improve eslint configuration

septs:


Jack-Works:
@septs I'd like to defer this change because there are two big refector is going on.

septs:
@Jack-Works OK

PR #2612, opened 2021-03-16
chore: migrate cli to typescript

septs:


packages/cli/build/main.ts
```diff
@@ -35,9 +38,12 @@ async function main(mode) {
     args.filter((x) => !x.startsWith('-')).forEach((target) => {
         command.push('--env', target)
         if (!knownTargets.includes(target)) {
-            throw new TypeError('Unknown target ' + target + '. Known targets: ' + knownTargets.join(','))
+            throw new TypeError(`Unknown target ${target}. Known targets: ${knownTargets}`)
```
septs, line 41: `knownTargets`, javascript array `toString` default behavior is use comma join string.

```
> foo = [1, 2, 3]
> `${foo}`
< "1,2,3" 
```

.github/workflows/linters.yml
```diff
@@ -21,19 +21,26 @@ jobs:
...
           name: eslint
           path: ./reports/junit
+  prettier:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: actions/setup-node@v2-beta
+      - run: npx prettier --list-different .
```
Jack-Works, line 43: I like this change

packages/cli/build/dev.ts
```diff
@@ -0,0 +1,16 @@
...
+import onMain from './main'
+import { build, dev } from './typescript'
+
+async function main() {
+    await build().catch(console.error)
+    dev()
+    if (process.argv[2] === '--') {
+        spawnSync(process.argv.slice(3).join(' '), { shell: true, stdio: 'inherit' })
+    } else {
+        onMain('dev')
```
Jack-Works, line 12: Suggestion: If the `onMain` or `spwanSync` dies, quit the process with that error

packages/cli/build/process-lock.ts
```diff
@@ -0,0 +1,24 @@
+import path from 'path'
+import { check, lock } from 'proper-lockfile'
```
Jack-Works, line 2: Actually this library doesn't always work well

packages/cli/build/main.ts
```diff
@@ -35,9 +38,12 @@ async function main(mode) {
...
         }
     })
-    spawn('npx', ['webpack', ...command], { stdio: 'inherit', shell: true, cwd: resolve(__dirname, '../maskbook/') })
+    return spawnSync('npx', ['webpack', ...command], {
```
Jack-Works, line 44: The child_process.spawnSync() method is generally identical to child_process.spawn() with the exception that the function will not return until the child process has fully closed.

I think this is not correct. It will block other daemons to do their tasks

septs: OK, i need handle exit code logic,
if non-zero them exit program

packages/cli/build/typescript.ts
```diff
@@ -0,0 +1,34 @@
...
+    shell: true,
+}
+
+export async function dev() {
+    lock: for await (const lock of makeLockSequence()) {
+        if (await lock()) break lock
+    }
+    runCli({ config, watch: true }, console.error)
+    // start a one-time "tsc -b" or parallel "tsc -b -w"
+    return spawn('npx', ['tsc', '-b', '-w'], options)
```
Jack-Works, line 20: TypeScript might die. Would you like to add code to restart tsc when it dies?

septs: yourself solve fix this problem?

Jack-Works: well not urgent.

Jack-Works (approved):

PR #2621, opened 2021-03-18
RFC: external plugins

Jack-Works:
See the rendered version at https://github.com/DimensionDev/Maskbook/blob/rfc/plugin/docs/rfc/000-External-plugin.md

### DIFF
--- a/docs/rfc/000-External-plugin.md
+++ b/docs/rfc/000-External-plugin.md
@@ -2,19 +2,19 @@
 
 ## Principles
 
-- It must be absolutely isolated with the Mask Network.
+- It must be absolutely isolated within the Mask Network extension.
 - Developers should be able to develop a working plugin without acquiring "licence" or any form of agreement from the Mask team.
 - It must not harm the data or property (wallet) safety.
 
 ## General design
 
-An external plugin should be an isolated web site, which means it should not be loaded / installed to the Mask Network.
+An external plugin should be an isolated website, which means it should not be loaded or installed onto the Mask Network.
 
-When Mask Network found a metadata of an external plugin, it will try to fetch the manifest and display the content in the payload. This process **must not** involves dynamic code execution.
+When Mask Network detects the metadata of an external plugin, it will try to fetch the manifest and display the content in the payload. This process **must not** involve dynamic code execution.
 
-When the user decided to interact with the external plugin, Mask Network will open a new popup window and inject some API to it.
+When the user decides to interact with the external plugin, Mask Network will open a popup window and inject some APIs to it.
 
-It is the same when user want to use the external plugin when composing.
+It is the same when user wants to use the external plugin when composing.
 
 ## User story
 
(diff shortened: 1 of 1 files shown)


Jack-Works:
Two changes:

21c60ab0131f2e5425b2f77efb398ea81f3d7eed to change the metadata info
6cb9812a1563c38618a276ab6973b6eca4962eae add the plugin badge in the composition dialog

docs/rfc/000-External-plugin.md
```diff
@@ -0,0 +1,202 @@
...
+      "method": "queryDetail",
+      "args": ["someVal", 0, "$.id"]
+    }
+  }
+}
+```
+
+### API in popups
+
+#### Mask.setMetadata(key: string, value: object)
```
septs, line 188: i want's need add an `namespace` parameter in the method

```suggestion
#### Mask.setMetadata(namespace: string, key: string, value: object)
```

identify different plugin.

docs/rfc/000-External-plugin.md
```diff
@@ -0,0 +1,202 @@
...
+      "key": "__locale:preview_description",
+      "map": {
+        "desc": "$.description",
+        "balance": "#.balance"
+      }
+    },
+    "action_text": "__locale:open_to_check",
+    "action_url": "./popup.html"
+  },
+  "vars": {
```
septs, line 175:
```suggestion
  "variables": {
```

docs/rfc/000-External-plugin.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Technical details
+
+### Plugin definition
+
+A plugin should be deployed on a static HTTPs URL, for example <https://example.com/my-plugin>, let's call it _base url_.
+
+It should provide a manifest file called "mask-plugin-manifest.json". 
+For the example above, it should be located at <https://example.com/my-plugin/mask-plugin-manifest.json>. 
+It should also be in JSONC (JSON with comment) format.
```
septs, line 64: may be need new JSON parser, https://mdn.io/JSON.parse not support JSONC
or use webpack pre-handler to JSON format.

docs/rfc/000-External-plugin.md
Jack-Works: why you need a symbol here?

docs/rfc/000-External-plugin.md
Jack-Works: this pattern is dropped. I'll investigate how this pattern is going to work when implementing

docs/rfc/000-External-plugin.md
septs: internal use?

docs/rfc/000-External-plugin.md
septs: OK

docs/rfc/000-External-plugin.md
Jack-Works:
```js
return x
        .split('\n')
        .filter((x) => !x.match(/^ +\/\//))
        .join('\n')
```

this is enough

docs/rfc/000-External-plugin.md
Jack-Works: this is part of our public API and it is going to send to other pages so you cannot transfer a symbol

amark:
@yisiliu @Jack-Works the way I read this seems me to think it is more for one off prompts and "cards"? Like, "approve a transaction" or something?

What I had told Yisi about a long time ago as an "app store" was going to be a lot more generic/powerful than this. The way to keep it secure tho is by never letting the app or extension or whatever actually touch/see the decrypted data per my [Secure Render](https://hackernoon.com/secure-rendering-a-new-browser-standard-for-user-privacy-dv783b9h ) design that is prototyped in Party (prevents F******k from reverse scraping your data). This actually gets very tricky if you aren't able to fully proxy/take-over the "browser" to reroute things, so there'd have to be varying level of trust as "app extensions"  slowly decrease over time from what they actually see, and this'd be fixed inbetween with permission prompts (as annoying as they might be, better safe than sorry).

Handling the permission prompts themselves, seems like an example territory of what you guys are proposing here? Or am I missing some more powerful abilities?

yisiliu:
@amark hey yeah thanks for your feedback! To us, we are facing a set of "issues" designing this, which are 1. database isolation 2. difficulty of "app" development 3. expressiveness over security policy bounded by the browser itself. We've thoroughly thought about the tradeoffs here and posted this RFC. 
Alternatively, what we are proposing is cutting the application off our extension by only displaying a "card" on Twitter instead of a fully function one. The actual "logic" is implemented in a separate web page so that we can separate the data in different indexdbs. This also sets off the limits of browser extensions.
Note that this is only one way of developing an app. You can still use our traditional way. We are not shutting the door. Instead, we are opening another one.

Jack-Works:
I'm sorry I don't quite understand the question you (@amark) have.
This RFC is intended to allow third-party plugins without any dynamic code execution in the Mask extension itself.
The brief model is the dApp model (like MetaMask and its `window.web3` or `window.ether` API).

Jack-Works:
Another thing to mention, the `The "payload_preview" part` has changed a lot but I've not updated it yet. I'm working on the prototype of external plugins to find a suitable way to do this.

Jack-Works:
https://github.com/DimensionDev/Maskbook/pull/2621/commits/249064d6842eeac71adc66c6adab2ef9da090359
Changed the manifest format and preview file to HTML+Web Component

Jack-Works:
I have complete the PoC demo in #2702. I'll update the RFC later to reflect the latest status quo.

septs:
@Jack-Works merge ready?

Jack-Works:
Maybe not. Let me modify it today

PR #2625, opened 2021-03-19
WIP: add multi-curve support in payload

yisiliu:


docs/rfc/000-Payload-v37.md
```diff
@@ -24,8 +24,11 @@ type PayloadAlpha37 = [
   // we can save more space
   authorNetwork: UTF8String | SocialNetworkEnum,
   authorID: UTF8String,
-  authorPublicKey: ArrayBuffer,
-  // authorPublicKeyCurve: ???,
+  //authorPublicKey: ArrayBuffer,
+  // array of base58 strings of newly generated ephemeral public keys
+  // indexed by SupportedCurves
+  authorEphemeralPublicKeyList: [String],       // length = Object.keys(SupportedCurves).length
```
Jack-Works, line 30: Why not ArrayBuffer (binary data)? (And do you mean `String[]`)

docs/rfc/000-Payload-v37.md
```diff
@@ -24,8 +24,11 @@ type PayloadAlpha37 = [
   // we can save more space
   authorNetwork: UTF8String | SocialNetworkEnum,
   authorID: UTF8String,
-  authorPublicKey: ArrayBuffer,
-  // authorPublicKeyCurve: ???,
+  //authorPublicKey: ArrayBuffer,
```
Jack-Works, line 27: Please leave this part. authorPublicKey is not part of Ephemeral encrytion process. It's persona discover

Jack-Works: And without `authorPublicKeyCurve` Mask cannot discover the persona correctly (don't know which curve on)

yisiliu: With `SupportedCurves` we can actually append a single byte next to the `authorPublicKey` indicating the curve. What do you think?

docs/rfc/000-Payload-v37.md
```diff
@@ -24,8 +24,11 @@ type PayloadAlpha37 = [
   // we can save more space
   authorNetwork: UTF8String | SocialNetworkEnum,
   authorID: UTF8String,
-  authorPublicKey: ArrayBuffer,
-  // authorPublicKeyCurve: ???,
+  authorPublicKey: ArrayBuffer,                 // format: [PublicKey, SupportedCurvesEnum]
+  // array of base58 strings of newly generated ephemeral public keys
+  // indexed by SupportedCurves
+  authorEphemeralPublicKeyList: String[],       // length = Object.keys(SupportedCurves).length
```
septs, line 30:
```suggestion
  authorEphemeralPublicKeyList: string[],       // length = Object.keys(SupportedCurves).length 
```

docs/rfc/000-Payload-v37.md
Jack-Works: Yes, a CurveEnum is good

docs/rfc/000-Payload-v37.md
```diff
@@ -24,8 +24,11 @@ type PayloadAlpha37 = [
   // we can save more space
   authorNetwork: UTF8String | SocialNetworkEnum,
   authorID: UTF8String,
-  authorPublicKey: ArrayBuffer,
-  // authorPublicKeyCurve: ???,
+  authorPublicKey: ArrayBuffer,                 // format: [PublicKey, SupportedCurvesEnum]
+  // array of base58 strings of newly generated ephemeral public keys
+  // indexed by SupportedCurves
+  authorEphemeralPublicKeyList: string[],       // length = Object.keys(SupportedCurves).length 
+  // authorPublicKeyCurve no longer necessary
```
Jack-Works, line 31:
```suggestion
  authorPublicKey: ArrayBuffer,
  authorPublicKeyCurve: SupportedCurvesEnum,
  authorEphemeralPublicKeyList: ArrayBuffer[],
```

docs/rfc/000-Payload-v37.md
```diff
@@ -24,8 +24,11 @@ type PayloadAlpha37 = [
   // we can save more space
   authorNetwork: UTF8String | SocialNetworkEnum,
   authorID: UTF8String,
-  authorPublicKey: ArrayBuffer,
-  // authorPublicKeyCurve: ???,
+  authorPublicKey: ArrayBuffer,                 // format: [PublicKey, SupportedCurvesEnum]
+  // array of base58 strings of newly generated ephemeral public keys
+  // indexed by SupportedCurves
+  authorEphemeralPublicKeyList: string[],       // length = Object.keys(SupportedCurves).length
```
septs, line 30:
```suggestion
  authorEphemeralPublicKeyList: SupportedCurve[],       // the list is unique?
```

docs/rfc/000-Payload-v37.md
```diff
@@ -43,6 +46,11 @@ enum SocialNetworkEnum {
   Facebook = 0,
   Twitter = 1,
 }
+enum SupportedCurves {
```
septs, line 49:
```suggestion
enum SupportedCurve {
```

docs/rfc/000-Payload-v37.md
yisiliu: This is not how it works. This list consists of a bunch of public keys not enums.

Jack-Works:
cc @yisiliu

docs/rfc/000-Payload-v37.md
yisiliu: why can't we encode the public key to `publicKey-curve`, e.g. `0x0Bd793EA8334A77b2BfD604DbAEdca11eA094306-secp256k1` or `pub17lgvvkwxlruqpfr6hzcy0wtg9s4nn7lpkwn0tz0kct9uzr624deqqdhsj8-ed25519`

Jack-Works: because no need to do this. we have a field for this so it can be parsed easier

yisiliu: But what I want to do is to support a more general encoding of a public key/address in a single string to be used somewhere else, e.g. in Tessercube, Twidere or any other supported systems.

Jack-Works: is it a standard to represent keys like this? And another thing to mention, instead of storing a string, I tend to store a binary (if is _is_ a binary)

yisiliu: Yes it can be found elsewhere like https://handbook.scuttlebutt.nz/concepts/identity. The only difference here is that they were using `.` as the separator but I think `-` is more URL-friendly. Binary sounds okay but since you are using msgpack, what's the difference then?

Jack-Works: You see, `0x0Bd793EA8334A77b2BfD604DbAEdca11eA094306` is representing an binary data. Converting it into string costs more spaces.

Jack-Works:
oops I merge the wrong branch on this PR

Jack-Works:
already included in the new RFC

PR #2647, opened 2021-03-23
feat: inject toolbox in navbar

zhouhanseng:
close https://github.com/DimensionDev/Maskbook/issues/2613


packages/maskbook/src/social-network/ui.ts
```diff
@@ -71,6 +71,7 @@ export async function activateSocialNetworkUI(): Promise<void> {
     ui.collecting.profilesCollector?.(signal)
     ui.injection.pageInspector?.(signal)
     if (Flags.toolbar_enabled) ui.injection.toolbar?.(signal)
+    ui.injection.toolBoxInNavBar?.(signal)
```
Jack-Works, line 74: Should we remove `ui.injection.toolbar`?

packages/maskbook/src/social-network/types.ts
```diff
@@ -103,6 +103,8 @@ export namespace SocialNetworkUI {
             postInspector?(signal: AbortSignal, current: PostInfo): void
             /** Inject a toolbar that displayed on the top of the page */
             toolbar?(signal: AbortSignal): void
+            /** Inject a tool box that displayed in the twitter navigation bar */
```
Jack-Works, line 106:
```suggestion
            /** Inject a tool box that displayed in the navigation bar of the SNS */
```

This is not only for Twitter.

packages/maskbook/src/social-network-adaptor/twitter.com/injection/Toolbox.tsx
```diff
@@ -0,0 +1,6 @@
+// import { injectToolboxDialogAtTwitter } from './ToolboxDialog'
+import { injectToolboxHintAtTwitter } from './ToolboxHint'
+
+export function injectToolBoxComposed(signal: AbortSignal) {
```
Jack-Works, line 4:
```suggestion
export function injectToolBox(signal: AbortSignal) {
```

packages/maskbook/src/social-network/ui.ts
zhouhanseng: Temporarily not.

Jack-Works: should it in a flag?

Jack-Works (approved):

packages/maskbook/src/social-network/ui.ts
zhouhanseng: oh, it requires flags rather than here but for some of its entries,  updated.

zhouhanseng: update

Jack-Works (approved):

PR #2707, opened 2021-03-30
fix: disallow other package manager lock-file

septs:
Avoid uploads by third-party contributors the non pnpm lock file


septs:
CC @Jack-Works

PR #2714, opened 2021-03-31
chore: disable react-hooks/exhaustive-deps rules

septs:


packages/.eslintrc.json
```diff
@@ -14,7 +14,7 @@
             "error",
             { "paths": [{ "name": "lodash", "message": "Please use lodash-es instead." }] }
         ],
-        "react-hooks/exhaustive-deps": "warn",
+        "react-hooks/exhaustive-deps": "off",
```
Jack-Works, line 17: so why not just remove that rule?

Jack-Works (changes requested):
~~~~

packages/.eslintrc.json
septs: current, we only can ignore all warning message on CI

Jack-Works (approved):

PR #2751, opened 2021-04-02
chore: remove unused imports

septs:


packages/.eslintrc.json
```diff
@@ -17,6 +17,7 @@
         "react-hooks/exhaustive-deps": "off",
         "react-hooks/rules-of-hooks": "error",
         "lodash/import-scope": ["error", "member"],
+        "unused-imports/no-unused-imports-ts": "error",
```
Jack-Works, line 20: Don't. That's annoying. `warning` is enough

Jack-Works (approved):

PR #2782, opened 2021-04-08
chore(docs): normalization bounty task

septs:
closes #

### DIFF
--- a/docs/i18n-guide.md
+++ b/docs/i18n-guide.md
@@ -1,0 +1,12 @@
+# Internationalization Guide (i18n guide)
+
+## Caveats
+
+1. We using [ISO 639-1][iso-639-1] two-letter code for locale directory name
+
+2. Add a new language new support
+
+   Please use [Add new locale request][request]
+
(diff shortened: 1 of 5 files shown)


guanbinrui (approved):

docs/i18n-guide.md
```diff
@@ -0,0 +1,14 @@
+# Internationalization Guide (i18n guide)
+
+## Caveats
+
+1. We using [ISO 639-1][iso-639-1] two-letter code for locale directory name
+
+   But `zh` as Traditional chinese
```
Jack-Works, line 7: Actually not. I select `zh` not `zh-hant` (or other) cause I don't know if `zh-hans` will use fallback to `zh-hant` instead of `en`

septs: plz you update this file.

Jack-Works: I searched for `zh` in ISO 639-1, it didn't mention that `zh` means `zh-hans` so it is okay to use `zh-hant` IMO. So remove this line is enough

Jack-Works (approved):

guanbinrui (approved):

PR #2827, opened 2021-04-14
feat: new concurrent mode helper

Jack-Works:
[jotai](https://github.com/pmndrs/jotai) is a ~concurrent-mode-ready~ library that supports async computation. I'm not intended to switch to jotai everywhere cause IMO React `useState` is good enough but I use it as an internal of our concurrent mode helper.

If you have read our current implementation (packages\shared\src\react\createGlobalState.ts and packages\shared\src\react\createResource.ts) you will find it's very complex.

Here I introduced a new API `createSuspended`. Here is how it used like

```ts
const useSomeAsyncTask = createSuspended(async (arg: string) => {
    return arg
})
function useComp() {
    const [data, update, isPending] = useSomeAsyncTask({ init: [''] })
    data.unwrap().toUpperCase()
    update('new text')
    isPending // same as useTransition
}
```

It can create a concurrent mode hook for async tasks. The hook needs an `init` argument to do the first time fetching and by the second time you need to call `update` (the 2nd argument) to trigger a suspended update.

The `data` returns in `Result<T, E>` the checked exception style to make sure you always checked the result. The third parameter is the same as the 2nd return of the React `useTransition` hook.

I don't have time to test this so I mark it as a draft for now. How do you think this API is designed? cc @zhouhanseng @nuanyang233 @guanbinrui

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -5,6 +5,7 @@
     "async-call-rpc": "^5.0.0",
     "bignumber.js": "^9.0.1",
     "immer": "^9.0.1",
+    "jotai": "^0.16.0",
     "ts-results": "^3.1.1",
     "typeson": "^6.0.0",
     "typeson-registry": "^1.0.0-alpha.38",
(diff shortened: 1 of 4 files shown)


Jack-Works:
tested. Jotai is not ready for it, I might have some misunderstanding. Related: https://github.com/pmndrs/jotai/issues/426

PR #2969, opened 2021-04-22
build: Mask Twitter Plugin Update

rob-lw:
+ removed vcent branding
+ reformatted text alignment
+ updated api
+ plugin component displays on particular tweet only


Jack-Works:
@rob-lw this pr contains merge conflicts with our main branch

septs (changes requested):
plz rebase master, thanks

packages/maskbook/src/plugins/VCent/define.tsx
```diff
@@ -15,6 +15,7 @@ export const VCentPluginDefine: PluginConfig = {
     postInspector: function Component(): JSX.Element | null {
         const tweetAddress = usePostInfoDetails('postID')
         if (!tweetAddress) return null
+        if (window.location.href.includes('v.cent.co')) return null
```
chejazi, line 18: Hi @rob-lw ! The goal of this change was that we aren't loading the component for every tweet in the feed, only the tweet that has been clicked on. Could we update to this?

`if (window.location.pathname.split('/')[3] != tweetAddress) return null`

rob-lw: Hi @chejazi. ah okay that is a misunderstanding on my part. I have updated to this.

packages/maskbook/src/plugins/VCent/constants.ts
```diff
@@ -1,6 +1,6 @@
 //#region plugin definitions
 export const PLUGIN_IDENTIFIER = 'com.maskbook.tweet'
 // proxy: https://v.cent.co/data/tweet-txn
-export const TWEET_BASE_URL = 'https://vcent-agent.r2d2.to/data/tweet-txn?tweetID='
+export const TWEET_BASE_URL = 'https://mod9gut7x5.execute-api.us-west-2.amazonaws.com/mask?tweetID='
```
Jack-Works, line 4: what is this URL? do you know @septs?

packages/maskbook/src/plugins/VCent/define.tsx
```diff
@@ -15,6 +15,7 @@ export const VCentPluginDefine: PluginConfig = {
     postInspector: function Component(): JSX.Element | null {
         const tweetAddress = usePostInfoDetails('postID')
         if (!tweetAddress) return null
+        if (window.location.pathname.split('/')[3] != tweetAddress) return null
```
Jack-Works, line 18: So this is only going to display on the detail page of a tweet? I think it might be a good approach to resolve #2840

packages/maskbook/src/plugins/VCent/constants.ts
septs: NO

packages/maskbook/src/plugins/VCent/define.tsx
Jack-Works: cc maybe @yisiliu or @guanbinrui?

packages/maskbook/src/plugins/VCent/constants.ts
chejazi: This is a proxy we set up using AWS API Gateway. v.cent.co doesn't include twitter.com in the CORS response header

packages/maskbook/src/plugins/VCent/define.tsx
chejazi: That was the intent of the change.

Jack-Works:
I think it might be ok on my view..

Tedko:
@Jack-Works @septs any updates ?

Jack-Works (approved):

Jack-Works (approved):

packages/maskbook/src/plugins/VCent/constants.ts
Tedko, line 4: need to test whether this api work or not

Tedko (approved):

Tedko:
this is almost done. lets deliver this this week @Jack-Works @septs @zhouhanseng after proper testing

PR #3007, opened 2021-04-25
feat: upgrade react-router to v6 in dashboard

### LINKED ISSUE #2989: [Demand] Upgrade react-router to v6 in dashboard
## Catalogue

- [ ] UI/UX
- [ ] Mask Core Feature
- [x] Mask Plugin Feature
- [ ] Other
  - /_List it here_/

The [migrating react router v5 to v6](https://github.com/ReactTraining/react-router/blob/dev/docs/advanced-guides/migrating-5-to-6.md) introduce many powerful upgrade in v6(relative routes and links, `<Route element>` ). We can upgrade it in the dashboard. @Jack-Works @relaxgo

nuanyang233:
closes #2989


packages/dashboard/src/components/DashboardFrame/Navigation.tsx
```diff
@@ -12,15 +12,16 @@ import {
...
 import { Link, LinkProps } from 'react-router-dom'
-import { Routes } from '../../pages/routes'
+import { RoutePaths } from '../../pages/routes'
 import { DashboardContext } from './context'
 import { MaskNotSquareIcon } from '@dimensiondev/icons'
 import { useDashboardI18N } from '../../locales'
 
 function ListItemLinkUnStyled({ nested, ...props }: LinkProps & ListItemProps & { nested?: boolean; to: string }) {
-    return <MuiListItem button component={Link} selected={!!useRouteMatch(props.to)} {...props} />
+    // @ts-ignore
```
nuanyang233, line 23: When upgrade react-router to v6, `component={Link}` will get a ts error
```
 Property 'component' does not exist on type 'IntrinsicAttributes & { button: true; } & { alignItems?: "flex-start" | "center" | undefined; autoFocus?: boolean | undefined; children?: ReactNode; ... 8 more ...; sx?: SxProps<...> | undefined; } & Omit<...> & CommonProps & Pick<...>'.
```
This is working fine in v5. I'm not sure which community to ask for help here. @Jack-Works

Jack-Works: Maybe both?

Jack-Works: And you cannot just ignore the ts error because it is a runtime behavior change.

nuanyang233:
https://github.com/ReactTraining/react-router/issues/7831

packages/dashboard/src/components/DashboardFrame/Navigation.tsx
Jack-Works: You can write a new component that wraps correctly warp & forwards ref of `component` prop to `as` prop. (https://github.com/ReactTraining/react-router/blob/dev/docs/advanced-guides/migrating-5-to-6.md#rename-link-component-to-link-as-and-navlink-exact-to-navlink-end)

nuanyang233: I did notice it here, but unfortunately `Link` doesn't have`as` prop

nuanyang233:
```ts
export declare const Link: React.ForwardRefExoticComponent<LinkProps & React.RefAttributes<HTMLAnchorElement>>;
export interface LinkProps extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'href'> {
    replace?: boolean;
    state?: State;
    to: To;
}
```

packages/dashboard/src/components/DashboardFrame/Navigation.tsx
```diff
@@ -5,27 +5,36 @@ import {
...
 
-function ListItemLinkUnStyled({ nested, ...props }: LinkProps & ListItemProps & { nested?: boolean; to: string }) {
-    return <MuiListItem button component={Link} selected={!!useRouteMatch(props.to)} {...props} />
+const ListItemLinkUnStyled = ({ to, ...props }: { to: string; nested?: boolean }) => {
+    const renderLink = useMemo(
+        () =>
+            forwardRef<HTMLAnchorElement | null, Omit<LinkProps, 'to'>>((linkProps, ref) => (
+                <Link to={to} ref={ref} {...linkProps} />
+            )),
+        [to],
```
Jack-Works, line 28: Hmm this part is strange to me... Why not define a new `Link` that internally renames `component` props (from mui) to `as` props (to react-route)?

nuanyang233: Link does not have `as` props in v6

nuanyang233:
It looks like react-router doesn't support <Link as={component} /> for now.  Close the pr first until the problem is solved.

https://github.com/ReactTraining/react-router/issues/7832

Jack-Works:
reopened but converted to draft.

Jack-Works:
Will you continue on this before we have more routes? @nuanyang233

nuanyang233:
> Will you continue on this before we have more routes? @nuanyang233

hmm, it looks like react-router doesn't solve the previously mentioned problem. I'll keep trying to replace it.

nuanyang233:
Sorry, I re-tested and the problem is not that `react-router` is missing `as` prop. Rather, it's a type issue with Material-ui When <ListItem button />,  this is the reason:

image

However, Material-ui provides <ListItemButton /> in version 36 to solve this problem
https://github.com/mui-org/material-ui/pull/26446

But they missing export it in version 36... https://github.com/mui-org/material-ui/issues/26660

maybe we need to wait version 38😂

Jack-Works:
We have `patch-package` available. You can patch material-ui to make things work. We're not going to upgrade to v35 recently because v35 needs to rewrite the import of `makeStyles` and I don't have the time budget for that.

Jack-Works:
@nuanyang233 I see the patch file is a little big. Is all that content necessary? Is it possible to do via a smaller (minimal) patch? That will be easier when we're going to upgrade.

Jack-Works (approved):

PR #3043, opened 2021-04-28
fix: fails to found PostInfo from @dimensiondev/mask-plugin-infra

UncleBill:
Fixes a problem as:

image

### DIFF
--- a/packages/maskbook/src/social-network/PostInfo.ts
+++ b/packages/maskbook/src/social-network/PostInfo.ts
@@ -1,5 +1,5 @@
 import { PostInfo } from '@dimensiondev/mask-plugin-infra'
-export { PostInfo } from '@dimensiondev/mask-plugin-infra'
+export { PostInfo }
 
 export const emptyPostInfo: PostInfo = new (class extends PostInfo {
     commentBoxSelector = undefined


packages/maskbook/src/social-network/PostInfo.ts
```diff
@@ -1,5 +1,5 @@
 import { PostInfo } from '@dimensiondev/mask-plugin-infra'
-export { PostInfo } from '@dimensiondev/mask-plugin-infra'
+export { PostInfo }
```
Jack-Works, line 2: Well, so this is strange... I think they're identical

Jack-Works (approved):

packages/maskbook/src/social-network/PostInfo.ts
UncleBill: Yep, it looks like https://github.com/babel/babel-loader/issues/603 (but it was [resolved in @babel/preset-typescript@7.12](https://github.com/babel/babel-loader/issues/603#issuecomment-737287260)).

guanbinrui (approved):

packages/maskbook/src/social-network/PostInfo.ts
Jack-Works: but we're not using babel

PR #3132, opened 2021-05-10
chore: simplify wallet formatter using

septs:
closes #


packages/shared/src/wallet/components.tsx
```diff
@@ -0,0 +1,71 @@
+import type { BigNumber } from 'bignumber.js'
+import { FC, Fragment } from 'react'
+import { formatBalance, formatChecksumAddress, formatCurrency, formatEthereumAddress, formatToken } from './formatter'
+
+interface FormattedBalanceProps {
```
Jack-Works, line 5: Those interface should also be exported otherwise use site cannot name those props.

Jack-Works (changes requested):
Other part LGTM

Jack-Works (approved):

zhouhanseng (approved):

guanbinrui (approved):

Jack-Works (approved):

PR #3160, opened 2021-05-12
fix: uesI18N rebinding every component-rendering and missing cleanup of unbindng

UncleBill:


packages/maskbook/src/utils/i18n-next-ui.ts
```diff
@@ -43,7 +43,10 @@ export function useI18N<NS extends keyof Namespaces = 'default'>(
         // eslint-disable-next-line react-hooks/rules-of-hooks
         const update = useUpdate()
         // eslint-disable-next-line react-hooks/rules-of-hooks
-        useEffect(() => document.addEventListener('i18n-hmr', update))
+        useEffect(() => {
+            document.addEventListener('i18n-hmr', update)
+            return () => document.removeEventListener('i18n-hmr', update)
+        }, [])
```
UncleBill, line 49: `useEffect(callback)` acts like `componentDidUpdate`, `useEffect(callback, [])` acts like `componentDidMount`

Jack-Works (approved):

PR #3178, opened 2021-05-14
refactor: hides organization details behind utils

UncleBill:
Just grep(vimgrep) `useI18N` and rewrite it and others exports from
utils I can see in `import { xxx } from 'path/to/utils'`

Rewriting all importees might introduce some unexpected changes, and
it's not friendly for code-reviewing, we might refactor rewrite other
importees progressively


packages/maskbook/tsconfig.json
```diff
@@ -2,7 +2,10 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "rootDirs": ["./src/", "./abis/"],
-    "outDir": "./dist/"
+    "outDir": "./dist/",
+    "paths": {
+      "@/*": ["./src/*"]
```
septs, line 7: @Jack-Works ?

Jack-Works:
The key point of this PR is going to introduce a path alias in the project. I'd like to against this change for the following reason.

1. The original use of path alias

The `paths` in the tsconfig was introduced to let the TypeScript compiler recognize the path that _can be_ recognized in the runtime, but TypeScript doesn't know that rule. TypeScript does not rewrite that for you (see https://github.com/microsoft/TypeScript/issues/5039#issuecomment-206451221). If the runtime does know about `@/` meaning, we should not use that.

"We're not running those codes directly, we're bundling it by Webpack" (a reasonable response to my argument) Actually this is not the case. The code might be running in Node.js for test/SSR. Things will break if you don't configure any of the following correctly:

- Webpack alias (for bundling)
- Cross-package bundling with path alias
- Node (test/SSR)

2. Toolchain complexity

Resolving path alias needs special knowledge of which TS tsconfig this file belongs to and a lot of work. We need to make sure the tools we're using support path alias, and if we want to make some tools on our own, we need to handle that too.
Give up the concise mapping from the file system to not well-known path rules to achieve convenience, I don't think it is worthy.


I'm strongly against path alias for the reasoning drafted above.

UncleBill:
Thanks, @Jack-Works, I would reconsider it.

This PR not only introduces path alias, but also combine all importees from `utils` folder by re-exports from `utils/index.ts`, somehow re-organizes import-exports, and hides organization details behide `utils`. We might make this change to other structured folders.

Jack-Works:
> but also combine all importees from utils folder by re-exports from utils/index.ts, somehow re-organizes import-exports, and hides organization details behide utils. We might make this change to other structured folders.

This part is good. I think we can start from this first.

Jack-Works (approved):

Issue #3122, opened 2021-05-08
[Bug] Rarible NFT display img glitch 


Tedko:
## Bug Info

<img width="2004" alt="Screen Shot 2021-05-09 at 04 55 10" src="https://user-images.githubusercontent.com/6687437/117553279-30bf9b80-b083-11eb-94aa-901888facf12.png">
<img width="2004" alt="Screen Shot 2021-05-09 at 04 56 08" src="https://user-images.githubusercontent.com/6687437/117553281-31f0c880-b083-11eb-96dd-f71d38322959.png">

### How to reproduce?

1. goto https://twitter.com/PunksTrucks/status/1388453137355362307?s=20
2. try it out - switch to rairble 
3.

### Mask Version:

- [ ] (Browser) Installed from the extension store (version: )
- [ ] (App) Installed from the app store (version: )
- [ ] master branch
- [ ] Build of version/commit hash:

(Please write down build command/flags if there is anything special)

### System

 
- [ ] Windows 10
- [ ] Mac OS X Catalina
- [ ] Linux: 

### Platform/Browser

 
- [ ] Latest Chrome (stable channel)
- [ ] Latest Firefox (stable channel)
- [ ] Android 11
- [ ] iOS 14

PR #3204, opened 2021-05-19
fix(collectible): nft from raible

UncleBill:
image

### DIFF
--- a/packages/maskbook/src/plugins/Collectible/hooks/useAsset.ts
+++ b/packages/maskbook/src/plugins/Collectible/hooks/useAsset.ts
@@ -90,7 +90,8 @@ export function useAsset(provider: CollectibleProvider, token?: CollectibleToken
                     is_verified: false,
                     is_owner: false,
                     is_auction: false,
-                    image_url: raribleResponse.properties.imagePreview,
+                    image_url:
+                        raribleResponse.properties.imagePreview ?? toRaribleImage(raribleResponse.properties.image),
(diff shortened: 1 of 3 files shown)


packages/maskbook/src/plugins/Collectible/hooks/useAsset.ts
```diff
@@ -90,7 +90,7 @@ export function useAsset(provider: CollectibleProvider, token?: CollectibleToken
                     is_verified: false,
                     is_owner: false,
                     is_auction: false,
-                    image_url: raribleResponse.properties.imagePreview,
+                    image_url: toRaribleImage(raribleResponse.properties.image),
```
nuanyang233, line 93: Please test [the issue](https://github.com/DimensionDev/Maskbook/issues/3122). 
Sometimes rarible doesn't return the `image`

Jack-Works (approved):

packages/maskbook/src/plugins/Collectible/hooks/useAsset.ts
UncleBill: Sometimes no imagePreview either.

UncleBill: I changed to use `imagePreview`, if it's not provided, then use `image`

nuanyang233 (approved):

Jack-Works (approved):

PR #3119, opened 2021-05-08
feat: add dHEDGE plugin

iRhonin:
This PR adds [dHEDGE](dhedge.org) plugin.

closes #2678
closes https://github.com/dhedge/dhedge-docs/issues/1

Demo Video: https://www.youtube.com/watch?v=i1xNTAi3o28

image

PR #3210, opened 2021-05-20
fix: locale key missing

septs:
closes #


Jack-Works:
Hmm, it seems like #3119 and #3034 conflict and create type errors on the master branch. cc @guanbinrui

PR #3234, opened 2021-05-26
fix: use useCallback in react components

Himself65:


nuanyang233:
🤔 It looks like these `usecallback` are not required to be used and will cause additional memory consumption.

Himself65:
> 🤔 It looks like these `usecallback` are not required to be used and will cause additional memory consumption.

Yes, you're right.

I test the code which uses `React.memo` and without, the previous code is faster than another(about 800ms vs. 1000ms).

I will open a new PR with that code.

packages/dashboard/src/pages/Settings/components/AncientPostsSetting.tsx
```diff
@@ -2,13 +2,14 @@ import type { ChangeEvent } from 'react'
...
+import { useCallback } from 'react'
 
 export default function AncientPostsSetting() {
     const checked = useAncientPostsCompatibilityMode()
 
-    const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const handleChange = useCallback((event: ChangeEvent<HTMLInputElement>) => {
         Services.Settings.setAncientPostsCompatibiltyMode(event.target.checked)
-    }
+    }, [])
```
Jack-Works, line 12: Actually this callback can be hoisted to the module level...

PR #3235, opened 2021-05-26
fix: version in decrypt

Himself65:


packages/maskbook/src/crypto/crypto-alpha-40.ts
```diff
@@ -245,8 +245,8 @@ export async function decryptMessage1ToNByMyself(info: {
     myLocalKey: AESJsonWebKey
     iv: string | ArrayBuffer
 }): Promise<[ArrayBuffer, AESJsonWebKey]> {
-    const { encryptedContent, myLocalKey, iv, encryptedAESKey } = info
-    const decryptedAESKey = await extractAESKeyInMessage(-40, encryptedAESKey, iv, myLocalKey)
+    const { version, encryptedContent, myLocalKey, iv, encryptedAESKey } = info
+    const decryptedAESKey = await extractAESKeyInMessage(version, encryptedAESKey, iv, myLocalKey)
```
Jack-Works, line 249: this file is the specific implementation of each version. IMO it should hard code the version number

PR #3236, opened 2021-05-26
perf: fix infinite tryRender invokes by setTimeout

### LINKED ISSUE #3152: [Bug] Performance issue - infinite tryRender invokes by setTimeout
### How to reproduce?

1. open https://twitter.com/differui
2. wait for UI and network stable
3. open chrome devtools, try `pause script execution`

image

### What happened?

image

### Mask Version:

- [x] (Browser) Installed from the extension store (version: 1.32.0)
- [ ] master branch

### System

 
- [ ] Windows 10
- [x] Mac OS X Catalina
- [ ] Linux: 

### Platform/Browser

 
- [ ] Android 11
- [ ] iOS 14

Jack-Works:
close #3152

First of all, I rename tryRender to other things so I fix the problem in the _literal meaning_. 🤣

Ok be serious, in this PR I investigate 3 fixes.

## Fix 1: better setTimeout

I share all instances to use 1 single `setTimeout` recursive call. That reduces the timers we registered.

Problem: It seems like there're still some nodes that cannot be connected to the document so the timer still runs forever.

## Fix 2: DOMNodeInserted

I observed the event of the ShadowRoot being connected to the DOM with `DOMNodeInserted` event.

Problem: This event is deprecated and it says this event harms the performance in other ways.

## Fix 3: MutationObserver

The problem of MutationObserver is we cannot use it to observe if the Node itself is connected. Thus by the solution given in https://github.com/whatwg/dom/issues/533#issuecomment-341880297, I have to listen to the whole document for DOM event changes.

Problem: Watching range is too wide. And because reason mentioned in Fix 1, it will still keep watching the whole document for any changes. I don't know if it really matters because our `MutatinoObserverWatcher` also watches the whole document.

We should determine which one to use.


packages/shared/src/ShadowRoot/createReactRootShadowed.tsx
```diff
@@ -31,6 +31,22 @@ export interface ReactRootShadowed {
     // do not name it as unmount otherwise it might be compatible with ReactDOM's Root interface.
     destory(): void
 }
+let observing = false
+const pendingMountPair = new Map<ShadowRoot, Function>()
+const mutationObserver = new MutationObserver(() => {
```
Himself65, line 36: Is it possible to use `mutationsList` in the first parameter, then filter with `pendingMountPair` that could be faster?

Jack-Works: I don't think either way will affect the performance. The key point is it observes the whole document changes.

Jack-Works:
hello, everyone plz vote on those solutions.

Himself65:
I like 3

Jack Works ***@***.***>于2021年6月10日 周四18:33写道：

Jack-Works:
@DimensionDev/dev-maskbook please vote

nuanyang233:
For 1st, we could probably set an upper limit

PR #3271, opened 2021-05-30
chore: install patch-package from npm source

septs:
closes #


Jack-Works:
@septs we're using an unreleased version. let's wait for that release.

PR #3279, opened 2021-06-01
ci: speedup ci by share build caches between build

Jack-Works:
Before (on commit c40cc5e8f62ccc30c74d58b673d9176174fb0cc3 which does not have a dependency changes):

- Type check spent 1m35s (2m52s in total) to run
- Webpack spent 5m52s (7m1s in total) to run

After (when cache hits):

- Type check spent 14s (3m23s in total) to run
- Webpack spent 2m22s (6m31s in total) to run

Note: I only optimized for Github Actions. Not for CircleCI or Netlify (maybe future)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,13 +23,18 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: "**/dist/"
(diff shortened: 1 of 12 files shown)


Jack-Works:
There is some progress on the build time but not too much. I guess it's an upstream bug and reported it. https://github.com/actions/cache/issues/597

PR #3313, opened 2021-06-10
chore: dns prefetch

septs:
closes #

### DIFF
--- a/packages/dashboard/public/index.html
+++ b/packages/dashboard/public/index.html
@@ -4,16 +4,6 @@
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <title>Mask Network</title>
-        <link rel="dns-prefetch" href="https://files.r2d2.to" />
-        <link rel="dns-prefetch" href="https://gun.r2d2.to" />
-        <link rel="dns-prefetch" href="https://service.r2d2.to" />
-        <link rel="dns-prefetch" href="https://coinmarketcap-agent.r2d2.to" />
-        <link rel="dns-prefetch" href="https://opensea-agent.r2d2.to" />
-        <link rel="dns-prefetch" href="https://vcent-agent.r2d2.to" />
-        <link rel="dns-prefetch" href="https://api.coingecko.com" />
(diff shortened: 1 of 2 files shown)


zhouhanseng (approved):

packages/dashboard/public/index.html
```diff
@@ -4,6 +4,16 @@
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <title>Mask Network</title>
+        <link rel="dns-prefetch" href="//files.r2d2.to" />
```
Jack-Works, line 7: Why you're not including `https:`?

Jack-Works (changes requested):

guanbinrui (approved):

packages/dashboard/public/index.html
Jack-Works: Our extension runs on `chrome-extension:` protocol.

Jack-Works (dismissed):

packages/dashboard/public/index.html
```diff
@@ -4,6 +4,16 @@
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <title>Mask Network</title>
+        <link rel="dns-prefetch" href="https://files.r2d2.to" />
```
Jack-Works, line 7: No need to prefetch in the dashboard. Because all request will be sent by the background page and it will do the dns-prefetch

septs: where?

Jack-Works: This page is used by the new dashboard. To make the new dashboard works, the user must have Mask Network installed. And you have already contained those DNS prefetches in the background page so those URLs are already prefetched.

Jack-Works:
Why close this? @septs

septs:
> Why close this? @septs

You don't think this improve is need

Jack-Works:
I think it's useful, but should only added in `packages/maskbook/miscs/template.html`

Jack-Works (approved):

NikoManiac (approved):

PR #3320, opened 2021-06-10
fix: remove duplicated code

Himself65:
closes #

### DIFF
--- a/packages/cli/locale-kit/locale-kit.ts
+++ b/packages/cli/locale-kit/locale-kit.ts
@@ -9,20 +9,8 @@ import {
     LOCALE_NAMES,
     readMessages,
     writeMessages,
+    findAllUnsyncedLocales,
 } from './utils'
-
-async function findAllUnsyncedLocales(locales = without(LOCALE_NAMES, 'en')) {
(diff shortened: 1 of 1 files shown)


packages/cli/locale-kit/locale-kit.ts
```diff
@@ -9,21 +9,9 @@ import {
     LOCALE_NAMES,
     readMessages,
     writeMessages,
+    findAllUnsyncedLocales,
```
UncleBill, line 12: It would be better to remove `findAllUnsyncedLocales` in `utils.js`. Anyway `functon findAllUnsyncedLocales()` should be alongside with `function removeAllUnusedKeys()`

Jack-Works (approved):

packages/cli/locale-kit/locale-kit.ts
septs: https://github.com/DimensionDev/Maskbook/blob/f47876ec70a305c25f7fe17b5c197c2f8ce3a071/packages/cli/locale-kit/locale-kit.ts#L75-L85

septs (approved):

PR #3333, opened 2021-06-11
fix: incorrect import path

nuanyang233:


guanbinrui (approved):

packages/icons/general/Search.tsx
```diff
@@ -1,5 +1,5 @@
 import { createIcon } from '../utils'
-import { MaskColorVar } from '@dimensiondev/maskbook-theme/constants'
+import { MaskColorVar } from '@dimensiondev/maskbook-theme'
```
Jack-Works, line 2: No this change will cause Netlify to fail to build. Please make sure pick a98674f4f7686cd28fcdb0561c8f30af9b3fe070 and 46f091d51d84c83510867070cf782134f078ef2f, then restore this change. cc @guanbinrui @nuanyang233

PR #3407, opened 2021-06-18
fix: listening transaciton event once to prevent memory leak

UncleBill:
closes #


Jack-Works (dismissed):

Jack-Works (approved):

Jack-Works (approved):

UncleBill:
web3.js has mechanism to automatically unsubscribe event, like

https://github.com/ChainSafe/web3.js/blob/bd6a9d86bbb9dc6e1fbe74216f9d7d44d7ba2b47/packages/web3-core-method/src/index.js#L342

so, closing.

PR #3506, opened 2021-06-25
fix: creation button can't be enabled in ImportWallet dialog

UncleBill:

### DIFF
--- a/packages/maskbook/src/utils/utils.ts
+++ b/packages/maskbook/src/utils/utils.ts
@@ -200,7 +200,7 @@ export function assert(x: any, ...args: any): asserts x {
 }
 
 export function checkInputLengthExceed(name: string) {
-    return Array.from(name).length >= WALLET_OR_PERSONA_NAME_MAX_LEN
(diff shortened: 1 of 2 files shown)


guanbinrui (approved):

septs (approved):

packages/maskbook/src/utils/utils.ts
```diff
@@ -200,7 +200,7 @@ export function assert(x: any, ...args: any): asserts x {
 }
 
 export function checkInputLengthExceed(name: string) {
-    return Array.from(name).length >= WALLET_OR_PERSONA_NAME_MAX_LEN
+    return name.length >= WALLET_OR_PERSONA_NAME_MAX_LEN
```
septs, line 203:
```
'💊'.length
2
Array.from('💊').length
1
```

UncleBill: I haven't learned deeply about emoji (An emoji is a glyph which may contains multiple unicode scalars), but I think we can follow twitter and browser's behavior.

image

image

https://codesandbox.io/s/jolly-tdd-rqn55?file=/src/App.js

PR #3508, opened 2021-06-25
feat: migrate file service

septs:
closes #


packages/maskbook/src/plugins/FileService/SNSAdaptor/components/Uploading.tsx
```diff
@@ -1,13 +1,13 @@
...
 import { File } from 'react-feather'
 import { useHistory, useLocation } from 'react-router'
 import { useAsync } from 'react-use'
-import { useI18N, timeout } from '../../../utils'
-import { PluginFileServiceRPC, PluginFileServiceRPCGenerator } from '../utils'
-import { FileRouter } from '../constants'
+import { timeout, useI18N } from '../../../../utils'
+import { FileRouter } from '../../constants'
+import type { FileInfo } from '../../types'
+import { PluginFileServiceRPC, PluginFileServiceRPCGenerator } from '../../utils'
```
Jack-Works, line 9: Please add a `Worker` entry for this plugin and import this file, otherwise, the RPC won't respond because it isn't started in the background page.

packages/maskbook/src/plugins/FileService/database.ts
```diff
@@ -1,12 +1,12 @@
 import { createPluginDatabase } from '../../database/Plugin/wrap-plugin-database'
```
Jack-Works, line 1: This file should be in ./Worker because it's only started in the Worker entry of this plugin.

packages/maskbook/src/plugins/FileService/Worker/index.ts
```diff
@@ -0,0 +1,11 @@
+import type { Plugin } from '@masknet/plugin-infra'
+import { base } from '../base'
+import './rpc'
+import './service'
```
Jack-Works, line 4: No need to add this, because ./service is lazy loaded in the ./rpc

packages/maskbook/src/plugins/FileService/define.tsx
```diff
@@ -8,11 +8,13 @@ import type { TypedMessage } from '../../protocols/typed-message'
```
Jack-Works, line 8: Would you like to rename this file to `helpers` or something like that?

septs:
@Jack-Works i want's split the plugin (file service) to subpackage (`@masknet/plugin-file-service`)

but

https://github.com/DimensionDev/Maskbook/blob/a76007ad99ff1acf8b25ec896199bc8606fcba61/packages/maskbook/src/plugins/FileService/Worker/database.ts#L1-L2

i don't know what to do

Jack-Works (approved):

zhouhanseng (approved):

guanbinrui (approved):

Issue #3414, opened 2021-06-18
[Demand] Find a tech solution to do lazy loading list


zhouhanseng:
## Catalogue

- [x] UI/UX
- [ ] Mask Core Feature
- [ ] Mask Plugin Feature
- [ ] Security/Cryptography Feature
- [ ] Other
  - /_List it here_/

## Part 1: Introduction and Background
We use [thegraph api](https://thegraph.com/docs/graphql-api#pagination) to fetch historical red packets and ITOs. It is unsuitable to fetch all red packets by one query over time. It‘s time to use lazy loading list (query `skip: 200 first: 100`)

## Part 2: Problem to Solve
There's a limitation, [thegraph api](https://thegraph.com/docs/graphql-api#pagination) will not tell you the total amounts of entities of one query, which means we can't calculate out the number of paginations in advance to use [Pagination Component](https://material-ui.com/components/pagination/).

## Part 3: Expected Behavior
Fetch 50 or 100 entities by each query. It will send a new query to fetch more data as soon as scrolling to the bottom of the history list. If there's no data returns from the last query, then stop listen to the scroll event for a while.

PR #3524, opened 2021-06-29
feat: ITO lazy load history list

zhouhanseng:
closes https://github.com/DimensionDev/Maskbook/issues/3414

https://user-images.githubusercontent.com/63733714/123776126-943aac80-d901-11eb-81c8-ff05c37c9145.mov


packages/maskbook/src/plugins/ITO/UI/PoolList.tsx
```diff
@@ -34,7 +34,10 @@ export function PoolList(props: PoolListProps) {
     const { value: pools = [], loading, retry } = useAllPoolsAsSeller(account, page)
 
     const containerRef = useRef<HTMLDivElement>(null)
-    const addPage = useCallback(() => setPage(page + 1), [page])
+    const addPage = useCallback(() => {
+        console.log({ page })
```
septs, line 38: remove log?

packages/shared/src/hooks/useScrollBottomEvent.ts
```diff
@@ -0,0 +1,27 @@
+import { useLayoutEffect, RefObject, useCallback } from 'react'
+import { debounce } from 'lodash-es'
+
+export function useScrollBottomEvent(ref: RefObject<HTMLDivElement>, cb: () => void) {
```
Jack-Works, line 4: I recommend using https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver.

zhouhanseng: Append an element at the bottom of the container, as it appears trigger the event. Maybe this is a little bit complex than the current implementation.

AinuX:
Test passed with build (29769)

packages/maskbook/src/plugins/ITO/UI/PoolList.tsx
zhouhanseng: removed

septs (approved):

nuanyang233 (approved):

packages/shared/src/hooks/useScrollBottomEvent.ts
Jack-Works: To use IntersectionObserver, we also add a new element at the bottom and observe its distance to the viewport. It's better than watching `scroll` event and calculate the distance by yourself.

You can use this: https://github.com/streamich/react-use/blob/master/docs/useIntersection.md

PR #3527, opened 2021-06-30
chore: clean file reader

septs:
closes #

https://github.com/DimensionDev/kit/commit/886532bfaaf28e0eebc3461b2c1964a39c731319

### DIFF
--- a/packages/dashboard/src/components/FileUpload/index.tsx
+++ b/packages/dashboard/src/components/FileUpload/index.tsx
@@ -2,6 +2,7 @@ import { MaskColorVar } from '@masknet/theme'
 import { makeStyles, Typography } from '@material-ui/core'
 import { useEffect, useState } from 'react'
 import { File as FileIcon } from '@masknet/icons'
+import { toText } from '@dimensiondev/kit'
 
 const useStyles = makeStyles(() => ({
     root: {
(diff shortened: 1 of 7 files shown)


Jack-Works (approved):

septs:
@yanzhihong23 
i think, `@dimensiondev/kit` provide blob helper function, no need provide `readAsText` to user.

yanzhihong23:
> @yanzhihong23
> i think, `@dimensiondev/kit` provide blob helper function, no need provide `readAsText` to user.

I think `readAsText` flag is fine. It's the most common case here.

UncleBill (approved):

packages/dashboard/src/components/FileUpload/index.tsx
```diff
@@ -2,6 +2,7 @@ import { MaskColorVar } from '@masknet/theme'
 import { makeStyles, Typography } from '@material-ui/core'
 import { useEffect, useState } from 'react'
 import { File as FileIcon } from '@masknet/icons'
+import { toText } from '@dimensiondev/kit'
```
yanzhihong23, line 5: The function name `toText` is not clear enough, what's the source type? Add a namespace or function name like `blobToText` would be better.

PR #3559, opened 2021-07-02
chore: plugin infra patches

Jack-Works:
This PR is simple and can be reviewed commit-by-commit


guanbinrui (approved):

packages/plugin-infra/src/types.ts
```diff
@@ -225,8 +234,29 @@ export namespace Plugin.Dashboard {
...
+        onBackup(): Promise<Option<unknown>>
+        /**
+         * This function will be called when the user try to restore a backup file,
+         * and there is some data associated with this plugin.
+         *
+         * @param data The serialized backup content previously returned by `onBackup`.
+         * You MUST treat the data as untrustful content because it can be modified by the user.
+         */
+        onRestore(data: unknown): Promise<Result<void, Error>>
+    }
```
septs, line 259:
```suggestion
    export interface BackupHandler<T> {
        /**
         * This function will be called when user try to generate a new backup.
         * The return value will contribute to the backup file.
         *
         * If it returns a None, it will not contributes to the backup file.
         *
         * If it returns a Some<T>, T will be serialized by JSON.stringify and added into the backup file.
         */
        onBackup(): Promise<Option<T>>
        /**
         * This function will be called when the user try to restore a backup file,
         * and there is some data associated with this plugin.
         *
         * @param data The serialized backup content previously returned by `onBackup`.
         * You MUST treat the data as untrustful content because it can be modified by the user.
         */
        onRestore(data: T): Promise<Result<void, Error>>
    }
```

Jack-Works: Using a type parameter will encourage plugins not to check their backup content via a type guard.

septs (approved):

PR #3568, opened 2021-07-02
refactor: rename constants to web3-constants

septs:
closes #


guanbinrui (approved):

packages/maskbook/tsconfig.json
```diff
@@ -12,7 +12,6 @@
   "references": [
     { "path": "../shared" },
     { "path": "../web3-shared" },
-    { "path": "../constants" },
```
Jack-Works, line 15: Why not this?
```suggestion
    { "path": "../web3-constants" },
```

septs: maskbook not need constants

https://github.com/DimensionDev/Maskbook/blob/e872ba1643a3e43f130088456551119eccc05a37/packages/maskbook/package.json#L21-L29

Jack-Works (approved):

PR #3643, opened 2021-07-10
chore: upgrade webextension-shim to fix mobile app issues

### LINKED ISSUE #3603: [Bug] Infinite loop during import JS as 'file' on iOS
### What happened?
On Mask Network iOS app, while decoding image format Mask encoded format payload, js files stuck on `resolveSocialNetworkWorker`, making image payloads can not be decoded.

The root cause seems to be that, during importing chunk js files, inside 'WebExtension-Shim`, `RunInProtocolScope` function has 
```
if (code.type === 'file')
    script.src = code.path;
```
This triggers the setter bound to `src` element. Inside the hooked setter, it invokes `RunInProtocolScope` again, leads to the infinite call loop.
Clipboard2
Clipboard

### Mask Version:

- [x] master branch

### System

 
- [ ] Windows 10
- [ ] Mac OS X Catalina
- [ ] Linux: 

### Platform/Browser

 
- [ ] Android 11
- [x] iOS 14

jk234ert:
closes #3603

### DIFF
--- a/packages/maskbook/package.json
+++ b/packages/maskbook/package.json
@@ -120,7 +120,7 @@
     "z-schema": "^5.0.1"
   },
   "devDependencies": {
-    "@dimensiondev/webextension-shim": "0.0.3-20201120022530-b318b89",
+    "@dimensiondev/webextension-shim": "0.0.3-20210708053012-57a254d",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.5.0-rc.0",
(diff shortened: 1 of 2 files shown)


UncleBill (approved):

Jack-Works (approved):

Jack-Works:
is that works? @jk234ert

jk234ert:
> does it work? @jk234ert

Yeah works as expected

PR #3651, opened 2021-07-12
fix: vcent plugin does not render tweets

lelenei:
closes #
MASK-121


guanbinrui (approved):

guanbinrui:
image

Jack-Works:
Will it cause #2840 again?

PR #3682, opened 2021-07-14
fix(build): watcher of webpack watches too many files including output files and files in node_modules

### LINKED ISSUE #3672: [DX problem] webpack keeps compiling repeatedly without file changes
Start it from my terminal without any editor open.

image

~~I guess its' because the `tsc` process is running, and it compiles `ts` into `js`, then triggers webpack to recompile.~~

UncleBill:
closes #3672

### DIFF
--- a/packages/maskbook/webpack.config.ts
+++ b/packages/maskbook/webpack.config.ts
@@ -14,7 +14,6 @@ import type { Configuration as DevServerConfiguration } from 'webpack-dev-server
 //#region Development plugins
 import ReactRefreshWebpackPlugin from '@pmmmwh/react-refresh-webpack-plugin'
 import ReactRefreshTypeScriptTransformer from 'react-refresh-typescript'
-import WatchMissingModulesPlugin from 'react-dev-utils/WatchMissingNodeModulesPlugin'
 import NotifierPlugin from 'webpack-notifier'
 //#endregion
 //#region Other plugins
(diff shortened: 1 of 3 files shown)


packages/maskbook/webpack.config.ts
```diff
@@ -48,6 +48,9 @@ function EnvironmentPluginNoCache(def: Record<string, any>) {
     }
     return new DefinePlugin(next)
 }
+const watchOptions = {
+    ignored: /\bnode_modules\b/,
```
Jack-Works, line 52: So... Who is changing node_modules?

UncleBill:
> So... Who is changing node_modules?

@Jack-Works  Maybe us.

Let's checkout the comparison of with and without ignoring:

| without ignoring | with ignoring |
|---|---|
| image | image |

We can see the watcher without ignoring watches `packages/maskbook/node_modules`, and it watches 49776 files!

Jack-Works:
Ok so please also remove `WatchMissingModulesPlugin` because I guess it does not work with node_modules not watched.

UncleBill:
> Ok so please also remove WatchMissingModulesPlugin because I guess it does not work with node_modules not watched.

Done.

Jack-Works (approved):

Jack-Works (approved):

nuanyang233 (approved):

PR #3735, opened 2021-07-22
refactor: let SearchIcon inherit color from style context

UncleBill:
also remove @masknet/theme from icons' dependencies

### DIFF
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -9,9 +9,6 @@
     "start": "snowpack dev",
     "build": "node --es-module-specifier-resolution=node ./dist/utils/ssr.js"
   },
-  "dependencies": {
-    "@masknet/theme": "workspace:*"
-  },
   "devDependencies": {
     "snowpack": "^3.8.1"
   }
--- a/packages/icons/general/Search.tsx
+++ b/packages/icons/general/Search.tsx
@@ -1,5 +1,4 @@
 import { createIcon } from '../utils'
-import { MaskColorVar } from '@masknet/theme/constants'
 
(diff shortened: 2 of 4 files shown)


septs (approved):

UncleBill:
I want to import an icon from `@masknet/icons` in `@masknet/theme`. `@/icons` already dependents `@/theme`, I don't want them to become a circular dependency to each other.

The reason I want an icon in `@/theme` is that a custom Snakbar I create in #3718 needs a spinning icon.

Reconsidering about the relationship between `@/theme` and `@/icons`, I think `@/theme` is a fundamental thing, it should be depended by any others, then we might need to move the components in `@/theme` to a new package to make `@/theme` more pure.

Would like to hear your opinion. @Jack-Works

septs (approved):

packages/icons/package.json
```diff
@@ -9,9 +9,6 @@
     "start": "snowpack dev",
     "build": "node --es-module-specifier-resolution=node ./dist/utils/ssr.js"
   },
-  "dependencies": {
-    "@masknet/theme": "workspace:*"
```
Jack-Works, line 13: If you remove this, please also remove it from tsconfig file.

UncleBill: I don't see any reference to `@masknet/theme` in `packages/icon/tsconfig.json`. What I miss?

guanbinrui (approved):

packages/icons/package.json
Jack-Works: Oh so that was a mistake when adding that dependency

PR #3744, opened 2021-07-23
fix: use `PersonaWithPrivateKey` type

Himself65:
https://github.com/DimensionDev/Maskbook/blob/160014e4cb9982f84c51f5ecfcb92e08e7bf2d9b/packages/maskbook/src/database/Persona/types.ts#L17-L19


packages/maskbook/src/database/Persona/Persona.db.ts
```diff
@@ -188,7 +189,7 @@ export async function queryPersonasDB(
     return records
 }
 
-export type PersonaRecordWithPrivateKey = PersonaRecord & Required<Pick<PersonaRecord, 'privateKey'>>
+export type PersonaRecordWithPrivateKey = PersonaWithPrivateKey
```
Jack-Works, line 192: but this type is unused. why not remove it?

PR #3770, opened 2021-07-27
fix: enhance type

Himself65:
I use these functions when doing `WebAuthn` support. But some of the types are incorrect. So, merge them would have benefit to the next future

### DIFF
--- a/packages/maskbook/src/modules/CryptoAlgorithm/interfaces/interface.ec.ts
+++ b/packages/maskbook/src/modules/CryptoAlgorithm/interfaces/interface.ec.ts
@@ -1,5 +1,4 @@
 import type { ShaName } from './interface'
-import type { AESName } from './interface.aes'
 import type { JsonWebKeyPair, EC_Public_JsonWebKey, EC_Private_JsonWebKey, AESJsonWebKey } from './utils'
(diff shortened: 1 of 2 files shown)


packages/maskbook/src/modules/CryptoAlgorithm/interfaces/interface.ec.ts
```diff
@@ -12,8 +12,8 @@ type Verify = (
 type DeriveAES = (
     priv: EC_Private_JsonWebKey,
     pub: EC_Public_JsonWebKey,
-    aes: AESName,
-    length: 256,
+    aes?: AESName,
```
Jack-Works, line 15: Why

Himself65: call it with only two parameter

septs: improve this function parameter to options pattern

```
type DeriveAES = (options: DeriveAESOptions) => PromiseLike<AESJsonWebKey>
```

Jack-Works: Why can't you call it with all parameters required?

PR #3779, opened 2021-07-27
refactor: cli rewrite

Jack-Works:
closes #


pnpm-workspace.yaml
```diff
@@ -2,7 +2,7 @@
 # webpack config.resolve.alias
 packages:
   - 'packages/empty'
-  - 'packages/cli'
```
septs, line 5: suggest, simplify file.

```
- packages/**
- packages/plugins/**
```

Jack-Works: I think we should be explicit here

septs (approved):

guanbinrui (approved):

PR #3798, opened 2021-07-29
refactor: compile contracts cli

zhouhanseng:
Fix: `Cannot find module '../cli/utils' or its corresponding type declarations.ts(2307)`


packages/scripts/package.json
```diff
@@ -1,10 +1,15 @@
...
   "name": "@masknet/cli",
   "private": true,
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
   "bin": {
     "build": "./src/bin/build.ts",
     "dev": "./src/bin/dev.ts"
   },
+  "scripts": {
+    "postinstall": "tsc"
```
zhouhanseng, line 11: Need to compile package after install (is like @masknet-shared compiled when `dev`), otherwise IDE will complaint`Cannot find module '@masknet/cli' or its corresponding type declarations.ts(2307)`

packages/scripts/src/codegen/guard.ts
```diff
@@ -12,8 +13,11 @@ function run(webpack: TaskFunction, cb: TaskFunctionCallback) {
     return shell.cwd(cwd)`${cmd}`
 }
 
-export const codegenGuardWatch = series(codegen, parallel(codegenWatch, run.bind(null, extensionWatch)))
-export const codegenGuard = series(codegen, run.bind(null, extension))
+export const codegenGuardWatch: Undertaker.TaskFunction = series(
```
zhouhanseng, line 16: `
The inferred type of 'codegenGuardWatch' cannot be named without a reference to '.pnpm/@types+undertaker@1.2.7/node_modules/@types/undertaker'. This is likely not portable. A type annotation is necessary.
`

packages/web3-contracts/compile-contracts.ts
```diff
@@ -35,11 +35,11 @@ async function main() {
     ])
 
     // format code
-    run(GENERATED_PATH, 'npx', 'prettier', '--write', '*')
+    shell`${GENERATED_PATH}, 'npx', 'prettier', '--write', '*'`
 
     // add to git stage
-    run(ABIS_PATH, 'git', 'add', '.')
-    run(GENERATED_PATH, 'git', 'add', '.')
+    shell`${ABIS_PATH}, 'git', 'add', '.'`
```
septs, line 41: safe?

septs: plz revert this modify.

zhouhanseng: @Jack-Works , why remove `run()` in https://github.com/DimensionDev/Maskbook/commit/e523a09a7682c8cc75af80a2a015ea38384f470e ?

Jack-Works: @zhouhanseng you should copy `run` from e523a09a7682c8cc75af80a2a015ea38384f470e into this folder. The content in `@masknet/cli` is not designed to be compiled or exported.

Jack-Works:
@zhouhanseng please revert the irrelated change, just copy `run` into the contract compiler. `@masknet/cli` must not be a dependency of another package.

packages/web3-constants/compile-constants.ts
```diff
@@ -1,6 +1,6 @@
 import { promises as fs } from 'fs'
 import path from 'path'
-import { run } from '../cli/utils'
+import { run } from '../web3-contracts/utils'
```
Jack-Works, line 3: Why not put it in `web3-constants`?

packages/scripts/src/codegen/guard.ts
Jack-Works: I'd prefer to revert this change because this package will not be type-checked.

zhouhanseng: okay

septs (approved):

Jack-Works (approved):

PR #3960, opened 2021-08-14
refactor: graphql query

hanszhoou:


packages/maskbook/src/plugins/ITO/Worker/apis/index.ts
```diff
@@ -48,17 +48,17 @@ const POOL_FIELDS = `
...
     const subgraphURL = getITOConstants(chainId ? chainId : currentChainIdSettings.value).SUBGRAPH_URL
     if (!subgraphURL) return null
-    const response = await fetch(subgraphURL, {
-        method: 'POST',
-        mode: 'cors',
-        body: stringify({ query }),
+
+    const apolloClient = new ApolloClient({
+        uri: subgraphURL,
+        cache: new InMemoryCache(),
```
zhouhanseng, line 57: We can use `client.readQuery({ query })` to to cache query results after fetching them.

packages/maskbook/src/plugins/ITO/Worker/apis/index.ts
```diff
@@ -90,33 +90,39 @@ export async function getTradeInfo(pid: string, trader: string) {
...
-            seller {
-                address
-            }
-            amount
-        }
-        destructInfos (where: { pool: "${pid.toLowerCase()}", seller: "${trader.toLowerCase()}" }) {
-            seller {
-                address
+    }>({
+        query: gql`
```
zhouhanseng, line 94: 1. Instead of injecting `${pid.toLowerCase()}` to string template directly, declare it in `variables` option.
2. `gql` template tag provides highlight makes code more readable.

PR #4039, opened 2021-08-19
feat: pick somthing from #3609

nuanyang233:
closes #


packages/maskbook/src/database/Persona/Persona.db.ts
```diff
@@ -64,8 +71,15 @@ const db = createDBAccessWithAsyncUpgrade<PersonaDB, Knowledge>(
                         }
                     }
                 }
+                async function v2_v3() {
+                    db.createObjectStore('relations', { keyPath: ['linked', 'profile'] })
+                    transaction
+                        .objectStore('relations')
+                        .createIndex('linked, profile, favor', ['linked', 'profile', 'favor'], { unique: false })
```
Jack-Works, line 78: I think it is unique👀

guanbinrui (dismissed):

packages/maskbook/src/database/Persona/Persona.db.ts
```diff
@@ -477,6 +501,86 @@ export async function deleteProfileDB(id: ProfileIdentifier, t: ProfileTransacti
...
+
+    for await (const cursor of t.objectStore('relations').index('linked, profile, favor').iterate()) {
+        if (cursor.value.linked !== linked.toText()) continue
+        if (
+            firstRecord &&
+            options.after &&
+            options.after.linked.toText() !== cursor?.value.linked &&
+            options.after.profile.toText() !== cursor?.value.profile
+        ) {
+            cursor.continue([options.after.linked, options.after.profile, options.after.favor].join(', '))
```
Jack-Works, line 539: Why did you join the index? I guess the index is an array

nuanyang233: yes

Jack-Works: So I think the parameter of continue should be an array

guanbinrui (approved):

Jack-Works (approved):

NikoManiac (approved):

PR #4066, opened 2021-08-23
feat: update connectProfile to directly use `ProfileIdentifier` as param

jk234ert:
Since `SNSAdaptor_getCurrentDetectedProfile` returns `ProfileIdentifier` as the result, the `persona_connectProfile` should also accept `ProfileIdentifier` as the input param

### DIFF
--- a/packages/maskbook/src/utils/native-rpc/Web.ts
+++ b/packages/maskbook/src/utils/native-rpc/Web.ts
@@ -11,7 +11,8 @@ import { WalletMessages } from '@masknet/plugin-wallet'
 import { WalletRPC } from '../../plugins/Wallet/messages'
 import { ProviderType } from '@masknet/web3-shared'
 
-const stringToIdentifier = (str: string) => Identifier.fromString(str, ECKeyIdentifier).unwrap()
+const stringToPersonaIdentifier = (str: string) => Identifier.fromString(str, ECKeyIdentifier).unwrap()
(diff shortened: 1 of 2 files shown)


guanbinrui (approved):

yanzhihong23 (approved):

packages/maskbook/src/utils/native-rpc/Web.ts
```diff
@@ -141,9 +141,11 @@ export const MaskNetworkAPI: MaskNetworkAPIs = {
 
         if (!result) throw new Error('invalid base64')
     },
-    persona_connectProfile: async ({ network, profileUsername, personaIdentifier }) => {
+    persona_connectProfile: async ({ profileIdentifier, personaIdentifier }) => {
+        const profileId = ProfileIdentifier.fromString(profileIdentifier)
```
Jack-Works, line 145:
```suggestion
        const profileId = ProfileIdentifier.fromString<ProfileIdentifier>(profileIdentifier, ProfileIdentifier)
```

packages/maskbook/src/utils/native-rpc/Web.ts
```diff
@@ -141,9 +141,11 @@ export const MaskNetworkAPI: MaskNetworkAPIs = {
...
+        if (!(profileId instanceof ProfileIdentifier)) throw new Error('invalid identifier')
```
Jack-Works, line 146: And remove this line

PR #4228, opened 2021-09-04
feat: thegraph failover

zhouhanseng:
Fetch from the chain with the block number(+1) of the latest record returned by thegraph. If there's no record returned by thegraph, then fetch with the block that the contract deployed.


packages/web3-shared/src/constants/index.ts
```diff
@@ -60,6 +61,8 @@ export const useSpaceStationGalaxyConstants = hookTransform(getSpaceStationGalax
 export const getOpenseaAPIConstants = transform(OpenseaAPI)
 export const useOpenseaAPIConstants = hookTransform(getOpenseaAPIConstants)
 
+export const getChainConstants = transform(Chain)
+export const useChainConstants = hookTransform(getChainConstants)
```
septs, line 65:
```suggestion
export const useChainConstants = hookTransform(getChainConstants)

```

packages/maskbook/src/plugins/ITO/schema-v1.json
```diff
@@ -119,19 +119,6 @@
         "seller": {
             "$ref": "#/definitions/trader"
         },
-        "buyers": {
```
zhouhanseng, line 122: Remove the useless `buyers` property. Through test, it would not break the old payload since `additionalProperties` is `true`, and `buyers` is also removed form `required` list.

zhouhanseng:
@DimensionDev/qa

AinuX:
Checked

guanbinrui (approved):

PR #4283, opened 2021-09-09
fix: `useSubmit` should capture `whoAmI`

### LINKED ISSUE #4280: [Bug] Fail to encrypt new message on iOS
### How to reproduce?

1.On iOS, open twitter, click the compose button
2.Click Mask Network icon to display the Mask Composition Dialog
3.Type in arbitrary texts
4.Click "Encrypt" button

### What happened?
The dialog re-create. The webview's console displays log message: "TypeError: Not initialized yet"
After some investigation, I found that inside the `useSubmit` method, I use `console.log()` to print the value of `whoAmI`:
```typescript
export function useSubmit(onClose: () => void) {
    const { t } = useI18N()
    const whoAmI = useCurrentIdentity() // Correct value printed here

    const onRequestPost = useCallback(
        async (info: SubmitComposition) => {
            const { content, encode, target } = info

            const [encrypted, token] = await Services.Crypto.encryptTo(
                content,
                target === 'Everyone' ? [] : target.map((x) => x.identifier),
                whoAmI?.identifier ?? ProfileIdentifier.unknown, // whoAmI is **undefined**
                target === 'Everyone',
            )
```
No idea why the `whoAmI` turns to be `undefined` when user do submit the message to encrypt, it makes the `encryptTo` method throwing the error "TypeError: Not initialized yet"

### Mask Version:

- [ ] master branch

### System

 
- [ ] Windows 10
- [ ] Mac OS X Catalina
- [ ] Linux: 

### Platform/Browser

 
- [ ] Android 11
- [x] iOS 14

jk234ert:
closes #4280

### DIFF
--- a/packages/maskbook/src/components/CompositionDialog/useSubmit.ts
+++ b/packages/maskbook/src/components/CompositionDialog/useSubmit.ts
@@ -50,7 +50,7 @@ export function useSubmit(onClose: () => void) {
             if (target !== 'Everyone') Services.Crypto.publishPostAESKey(token)
             onClose()
         },
-        [whoAmI?.identifier.toText(), onClose],
+        [t, whoAmI?.identifier.toText(), onClose],
     )
     return onRequestPost
 }
--- a/packages/maskbook/src/components/CompositionDialog/useSubmit.ts
+++ b/packages/maskbook/src/components/CompositionDialog/useSubmit.ts
@@ -50,7 +50,7 @@ export function useSubmit(onClose: () => void) {
             if (target !== 'Everyone') Services.Crypto.publishPostAESKey(token)
             onClose()
         },
-        [t, onClose],
+        [whoAmI?.identifier.toText(), onClose],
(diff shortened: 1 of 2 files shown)


packages/maskbook/src/components/CompositionDialog/useSubmit.ts
```diff
@@ -50,7 +50,7 @@ export function useSubmit(onClose: () => void) {
             if (target !== 'Everyone') Services.Crypto.publishPostAESKey(token)
             onClose()
         },
-        [t, onClose],
+        [whoAmI?.identifier.toText(), onClose],
```
Jack-Works, line 53:
```suggestion
        [t, whoAmI?.identifier.toText(), onClose],
```

Jack-Works (approved):

Jack-Works (approved):

guanbinrui (approved):

PR #4287, opened 2021-09-09
refactor: rewrite loading button

Jack-Works:
closes #


UncleBill:
I think the type should be `refactor` instead of `feat`, to make sure it will be place in correct place of release note.

PR #4288, opened 2021-09-09
fix: get coin price should pass coinId instead of platform

UncleBill:


UncleBill:
`urlcat` encode `,` in parameter, that's not expected

https://runkit.com/embed/ywz7ebmh0rxz

nuanyang233 (approved):

Jack-Works (approved):

PR #4315, opened 2021-09-10
feat: password field in dashboard

Lanttcat:
closes #

### DIFF
--- a/packages/theme/src/Components/TextField/PasswordField.tsx
+++ b/packages/theme/src/Components/TextField/PasswordField.tsx
@@ -4,20 +4,20 @@ import { Visibility, VisibilityOff } from '@material-ui/icons'
 import { IconButton, InputAdornment } from '@material-ui/core'
 
 export const MaskPasswordTextField = memo<MaskTextFieldProps>((props) => {
-    const [isShowPassword, setIsShowPassword] = useState(false)
+    const [hidden, setHidden] = useState(false)
 
     return (
         <MaskTextField
             {...props}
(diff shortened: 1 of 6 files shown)


packages/theme/src/Components/TextField/index.tsx
```diff
@@ -48,7 +48,7 @@ const useStyles = makeStyles()((theme) => ({
     },
 }))
 
-type MaskTextFieldProps = Exclude<StandardTextFieldProps, 'variant'>
+export type MaskTextFieldProps = Exclude<StandardTextFieldProps, 'variant'>
```
septs, line 51: why add the `Mask` prefix string?

packages/theme/src/Components/TextField/PasswordField.tsx
```diff
@@ -0,0 +1,27 @@
+import { MaskTextField, MaskTextFieldProps } from '.'
+import { memo, useState } from 'react'
+import { Visibility, VisibilityOff } from '@material-ui/icons'
+import { IconButton, InputAdornment } from '@material-ui/core'
+
+export const MaskPasswordTextField = memo<MaskTextFieldProps>((props) => {
+    const [isShowPassword, setIsShowPassword] = useState(false)
```
septs, line 7:
```suggestion
    const [hidden, setHidden] = useState(false)
```

packages/theme/src/Components/TextField/index.tsx
NikoManiac: TextField is belong to material, Add `mask` can help distinguish.

PR #4343, opened 2021-09-15
chore: rp maintain style

zhouhanseng:
closes #


AinuX:
Checked

UncleBill (approved):
The abbreviation rp for `red packet` is very confusing. Better to use `red packet` instead

nuanyang233 (approved):

PR #4386, opened 2021-09-18
feat: redpacket thegraph failover

hanszhoou:
closes #


packages/maskbook/src/plugins/RedPacket/Worker/database.ts
```diff
@@ -21,23 +24,6 @@ export async function addRedPacket(record: RedPacketRecord) {
     return RedPacketDatabase.add(RedPacketRecordIntoDB(record))
 }
 
-export async function updateV1ToV2(history: RedPacketHistory) {
```
Jack-Works, line 24: why?

zhouhanseng:
@DimensionDev/qa

zhouhanseng:
@Jack-Works Remove `updateV1ToV2()`, since all v1 red packets had expired(unable to claim) long time ago, and the password only used for user to claim.

packages/maskbook/src/plugins/RedPacket/Worker/database.ts
Jack-Works: I guess you should delete old records if you don't want them anymore.

PR #4411, opened 2021-09-21
fix: ens network

zhouhanseng:
Error: ENS is not supported on network private at ENS.checkNetwork

https://github.com/ChainSafe/web3.js/issues/3787


guanbinrui (approved):

Tedko:
Let's do a hot fix thx guys

PR #4430, opened 2021-09-22
chore: replace `webpack-extension-manifest-plugin`

septs:
closes #

### DIFF
--- a/packages/maskbook/webpack.config.ts
+++ b/packages/maskbook/webpack.config.ts
@@ -13,7 +13,7 @@ import ReactRefreshWebpackPlugin from '@pmmmwh/react-refresh-webpack-plugin'
 import CopyPlugin from 'copy-webpack-plugin'
 import HTMLPlugin from 'html-webpack-plugin'
 import WebExtensionTarget from 'webpack-target-webextension'
-import ManifestPlugin from 'webpack-extension-manifest-plugin'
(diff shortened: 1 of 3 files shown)


packages/maskbook/public/manifest.webextension-shim.json
```diff
@@ -1,11 +0,0 @@
-{
```
Jack-Works, line 1: Don't remove this file.

packages/maskbook/webpack.config.ts
```diff
@@ -365,22 +367,54 @@ export default async function (cli_env: Record<string, boolean> = {}, argv: Argv
...
-        else if (target.runtimeEnv.build === 'insider') modifiers.nightly(manifest)
+                if (target.runtimeEnv.build === 'beta') modifiers.beta(manifest)
+                else if (target.runtimeEnv.build === 'insider') modifiers.nightly(manifest)
 
-        return new ManifestPlugin({ config: { base: manifest } })
+                return JSON.stringify(manifest, null, 4)
+            },
+        })
+    }
+    function emitManifestWebExtShimFile() {
```
septs, line 392: @Jack-Works replace method

Jack-Works: but this file is purely static. it's better to maintain if it is static.

guanbinrui (approved):

Jack-Works (approved):

PR #4471, opened 2021-09-26
fix: should change token when change network on transfer

NikoManiac:
closes #


packages/dashboard/src/pages/Wallets/components/Transfer/TransferERC20.tsx
```diff
@@ -58,6 +58,10 @@ export const TransferERC20 = memo<TransferERC20Props>(({ token }) => {
     const is1559Supported = useMemo(() => isEIP1559Supported(chainId), [chainId])
     const { gasNow } = useGasOptions()
 
+    useEffect(() => {
+        setToken(token)
```
UncleBill, line 62: To avoid confusing, rename `setToken` to `setSelectedToken`
```suggestion
        setToken(token)
```

guanbinrui (approved):

PR #4499, opened 2021-09-27
fix: make the custom snackbar as default

UncleBill:
| On light mode | On dark mode |
| -- | -- |
| image | image |

## Change anchorOrigin (from top center)

image

### DIFF
--- a/packages/theme/src/Components/Snackbar/index.tsx
+++ b/packages/theme/src/Components/Snackbar/index.tsx
@@ -1,4 +1,4 @@
-import { forwardRef, useRef, memo } from 'react'
+import { forwardRef, useRef, memo, useCallback } from 'react'
 import { keyframes } from 'tss-react'
 import {
     SnackbarProvider,
@@ -25,6 +25,7 @@ export { SnackbarProvider, useSnackbar } from 'notistack'
 export type { VariantType, OptionsObject, SnackbarKey } from 'notistack'
 
 const useStyles = makeStyles()((theme, _, createRef) => {
+    const { palette } = theme
     const spinningAnimationKeyFrames = keyframes`
 to {
   transform: rotate(360deg)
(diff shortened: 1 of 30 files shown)


packages/theme/src/Components/Snackbar/index.tsx
```diff
@@ -224,3 +239,25 @@ export const CustomSnackbarProvider = memo<SnackbarProviderProps>((props) => {
         />
     )
 })
+
+export function useShowCostomSnackbar() {
```
nuanyang233, line 243:
```suggestion
export function useShowCustomSnackbar() {
```

nuanyang233 (changes requested):

packages/theme/src/Components/Snackbar/index.tsx
```diff
@@ -224,3 +239,25 @@ export const CustomSnackbarProvider = memo<SnackbarProviderProps>((props) => {
...
+    const snackbar = useSnackbar()
+    const showSnackbar = useCallback(
+        (
+            message: string,
+            options: Partial<CustomSnackbarContentProps & { persist: boolean }> = {
+                variant: 'default',
```
nuanyang233, line 249: Perhaps in most scenarios, `success` is better as the default value

packages/theme/src/Components/Snackbar/index.tsx
```diff
@@ -224,3 +239,25 @@ export const CustomSnackbarProvider = memo<SnackbarProviderProps>((props) => {
         />
     )
 })
+
+export function useShowCostomSnackbar() {
+    const snackbar = useSnackbar()
+    const showSnackbar = useCallback(
```
nuanyang233, line 245:
```suggestion
   return useCallback(
```

packages/theme/src/Components/Snackbar/index.tsx
UncleBill: It's all based on the design.

packages/theme/src/Components/Snackbar/index.tsx
UncleBill: Fixed

Jack-Works (approved):

PR #4533, opened 2021-09-30
feat: revamp red packet plugin

UncleBill:
recall of https://github.com/DimensionDev/Maskbook/pull/3718

part of https://github.com/DimensionDev/Maskbook/issues/3654


packages/maskbook/src/plugins/RedPacket/SNSAdaptor/RedPacket/useStyles.ts
```diff
@@ -1,107 +1,141 @@
 import { makeStyles } from '@masknet/theme'
+import { keyframes } from 'tss-react'
```
Jack-Works, line 2: Please reexport keyframes from `@masknet/theme` instead of depends on `tss-react` directly.

guanbinrui (approved):

PR #4580, opened 2021-10-11
fix: popups ui

nuanyang233:
closes MASK-215


nuanyang233:
Wait for test @wenluomask

PR #4620, opened 2021-10-19
fix: rewrite DialogStacking and plugin open-on-start to fix unresponsive

Jack-Works:
rewrite the following components to fix the time-racing problems:

- DialogStacking
- Composition dialog open-on-start


wenluomask:
Checked.

PR #4621, opened 2021-10-19
feat: dashboard change network improve

Lanttcat:
1. should open wallet states dialog before change network
2. should filter not verified token in debank asset api


wenluomask:
Need some UI ajustment and add more chinese translation.

PR #4777, opened 2021-11-07
feat: integrate solana chain

UncleBill:
based on #4723

image

image

image


packages/web3-shared/solana/tsconfig.json
```diff
@@ -2,8 +2,9 @@
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
     "rootDir": "./",
-    "outDir": "../dist/solana",
-    "tsBuildInfoFile": "../dist/solana.tsbuildinfo"
+    "outDir": "../dist/flow",
```
Jack-Works, line 5: why? name does not match the folder.

UncleBill: A mistake.

packages/plugins/Solana/src/locales/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import es_ES from './es-ES.json'
+import fa_IR from './fa-IR.json'
+import fr_FR from './fr-FR.json'
+import it_IT from './it-IT.json'
+import ja_JP from './ja-JP.json'
+import ko_KR from './ko-KR.json'
+import qya_AA from './qya-AA.json'
+import ru_RU from './ru-RU.json'
+import zh_CN from './zh-CN.json'
+import zh_TW from './zh-TW.json'
```
Jack-Works: We don't have the following languages currently. Please remove them and run `npx gulp sync-languages` again:

- es-ES
- fa-IR
- fr-FR
- it-IT
- ru-RU

packages/mask/.webpack/config.ts
```diff
@@ -120,10 +123,18 @@ export function createConfiguration(rawFlags: BuildFlags): Configuration {
                 { test: /(async-call|webextension).+\.js$/, enforce: 'pre', use: ['source-map-loader'] },
                 // TypeScript
                 {
-                    test: /\.tsx?$/,
+                    test: /\.(t|j)sx?$/,
+                    resolve: {
+                        fullySpecified: false, // disable the behaviour
```
Jack-Works: is it possible to not use this option? or limit it to a specific path by `test`.

UncleBill:
image

packages/plugins/Solana/src/locales/index.ts
UncleBill: Done.

packages/mask/.webpack/config.ts
UncleBill: No. without this, it will not work for `@solana/wallet-adapter-*`.

Jack-Works: what about this:

```js
{
    test: /@solana\/wallet-adapter/,
    resolve: { fullySpecified: false }
}
```

And you should report it to the upstream so we can remove it when they fixed that.

UncleBill: `@solana/wallet-adapter-*` need transpiling

> And you should report it to the upstream so we can remove it when they fixed that.

Some one already reported at https://github.com/solana-labs/wallet-adapter/issues/231

Jack-Works: OK, so please split it as a separate rule and don't touch our default rule

jordaaash: https://github.com/solana-labs/wallet-adapter/issues/231#issuecomment-1003465379

^ the packages have been updated, this issue may be fixed. The packages no longer require transpilation because they ship ESM and CJS versions now.

UncleBill: @jordansexton Wow, thanks.

UncleBill: Looks like it didn't fix. Still get same errors without those config updates.  @ jordansexton

image

jordaaash: I'm not sure what your build config is, but please first check out the starter projects in the WalletAdapter repo, since they are confirmed to be working. Please open an issue with a link to a reproduction if that doesn't help.

packages/dashboard/src/pages/Wallets/components/CollectibleList/index.tsx
```diff
@@ -49,20 +50,20 @@ export const CollectibleList = memo<CollectibleListProps>(({ selectedNetwork })
     const [renderData, setRenderData] = useState<Web3Plugin.NonFungibleToken[]>([])
 
     const {
-        value = { data: [], hasNextPage: false },
+        value = { data: EMPTY_LIST, hasNextPage: false },
```
UncleBill, line 52: Reduce re-rendering from 1000 times to about 10 times according to a test.

packages/dashboard/package.json
```diff
@@ -38,8 +38,8 @@
     "json-stable-stringify": "^1.0.1",
     "react-avatar-editor": "^12.0.0",
     "react-hook-form": "^7.22.5",
-    "react-router": "^6.0.0-beta.0",
-    "react-router-dom": "^6.0.0-beta.0",
+    "react-router": "6.0.0-beta.0",
```
UncleBill, line 41: https://github.com/DimensionDev/Maskbook/blob/d88e6eb5071cf69689cb54bfdc5b3231e8c82073/packages/dashboard/src/pages/routes.tsx#L21-L25 will fails at `react-router{,-dom}@6.1`, this is a temporary walkaround.

packages/dashboard/src/pages/Wallets/components/CollectibleList/index.tsx
```diff
@@ -44,25 +45,25 @@ export const CollectibleList = memo<CollectibleListProps>(({ selectedNetwork })
     const account = useAccount()
     const { Asset } = useWeb3PluginState()
     const network = useNetworkDescriptor()
-    const [loadingSize, setLoadingSize] = useState<number>()
-    const [loadingCollectible, setLoadingCollectible] = useState(true)
+    const [loadingSize, setLoadingSize] = useState(0)
```
septs, line 48: why initial value not is NaN?

```suggestion
    const [loadingSize, setLoadingSize] = useState(Number.NaN)
```

packages/plugins/Solana/src/UI/components/ProviderIconClickBait.tsx
```diff
@@ -0,0 +1,38 @@
...
+    }, [provider, onClick, onSubmit])
+
+    return (
+        <>
+            {isValidElement<object>(children)
+                ? cloneElement(children, {
+                      ...children.props,
+                      ...{
+                          onClick: onLogIn,
+                      },
```
septs:
```suggestion
                      onClick: onLogIn,
```

packages/dashboard/src/pages/Wallets/components/CollectibleList/index.tsx
UncleBill: Initialize with `0` is OK.

packages/plugins/Solana/src/UI/components/ProviderIconClickBait.tsx
UncleBill: Fixed

PR #4840, opened 2021-11-18
chore: web3 context related

hanszhoou:
closes #


packages/plugin-infra/src/web3/Context.tsx
```diff
@@ -81,11 +81,17 @@ export function usePluginIDContext() {
 
 const PluginsWeb3StateContext = createContainer(usePluginsWeb3State)
 
-export const usePluginsWeb3StateContext = PluginsWeb3StateContext.useContainer
```
zhouhanseng, line 84: `usePluginsWeb3State()` functions the same as `usePluginsWeb3StateContext()`

packages/plugin-infra/src/web3/Context.tsx
```diff
@@ -81,11 +81,17 @@ export function usePluginIDContext() {
...
+// usePluginWeb3Context is used to provide a set of API to interact with Web3.
+export function usePluginWeb3Context(expectedPluginID?: string) {
```
zhouhanseng, line 85: `usePluginWeb3Context()` functions this same as `useWeb3State()`, but more readable.

packages/plugin-infra/src/web3/useWeb3State.ts
```diff
@@ -1,7 +0,0 @@
-import { usePluginIDContext } from '.'
-import { useActivatedPluginWeb3State } from '..'
-
-export function useWeb3State(expectedPluginID?: string) {
```
zhouhanseng, line 4: There's a vague meaning between `useWeb3State()` and `usePluginWeb3StateContext()`

packages/shared-base/src/utils/compose.ts
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+export function compose(init: React.ReactNode, ...f: ((children: React.ReactNode) => JSX.Element)[]) {
```
Jack-Works, line 3: This MUST NOT be in the shared-base package because it imports React. Please move it to `shared` package.

packages/plugins/Flow/src/hooks/useFCL.ts
```diff
@@ -0,0 +1,8 @@
+import { useMemo } from 'react'
+import * as SDK from '@masknet/web3-shared-flow'
+
+export function useFCL() {
+    return useMemo(() => {
```
Jack-Works, line 5: You should use `useRef`

packages/shared-base/src/utils/compose.ts
```diff
@@ -0,0 +1,5 @@
...
+    return f.reduceRight((prev, curr) => curr(prev), React.createElement('div', '', init))
```
Jack-Works, line 4: Your version is different than the original version. The original version creates a Fragment instead of a div.

Jack-Works (changes requested):

PR #4891, opened 2021-11-23
chore(facebook): snackbar position

### LINKED ISSUE #4701: [Bug] Auto insert not work with Image payload
### How to reproduce?

1. Publish post with image payload
image

### What happened?
1. Image payload will not be pasted in mask compose automatically.
2. Popup dialog is out of screen

### Mask Version:

- [ ] master branch

### System

 
- [ ] Windows 10
- [x] Mac OS X Catalina
- [ ] Linux: 

### Platform/Browser

 
- [ ] Android 11
- [ ] iOS 14

hanszhoou:
closes #4701

image

### DIFF
--- a/packages/mask/src/UIRoot.tsx
+++ b/packages/mask/src/UIRoot.tsx
@@ -9,7 +9,7 @@ import i18nNextInstance from '../shared-ui/locales_legacy'
 import { Web3Context } from './web3/context'
 import { buildInfoMarkdown } from './extension/background-script/Jobs/PrintBuildFlags'
 import { activatedSocialNetworkUI } from './social-network'
-import { FACEBOOK_ID } from './social-network-adaptor/facebook.com/base'
+import { isFacebook } from './social-network-adaptor/facebook.com/base'
 import { pluginIDSettings } from './settings/settings'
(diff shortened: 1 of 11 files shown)


zhouhanseng:
This is a workaround patch. After `/packages/mask/src/social-network` becomes an independent package, we can solve this  elegantly.

packages/mask/src/UIRoot.tsx
```diff
@@ -33,7 +33,7 @@ function MaskThemeProvider({ children, baseline, useTheme }: MaskThemeProvider)
                 disableWindowBlurListener={false}
                 anchorOrigin={{ vertical: 'top', horizontal: 'right' }}
                 children={jsx}
-                isFacebook={activatedSocialNetworkUI.networkIdentifier === FACEBOOK_ID}
+                isfacebook={Boolean(isFacebook(activatedSocialNetworkUI)).toString()}
```
zhouhanseng, line 36: Browser complains that this DOM attribute should be a string, attribute name non-camelcase.

zhouhanseng:
repalce with #4910

PR #4900, opened 2021-11-23
fix(gun): register Gun early in background script

UncleBill:
Fix problem like this:

image

Root cause:

https://github.com/amark/gun/blob/1969d48e0fc20afd0c2fe2a3b4fefb98f745e7c3/lib/store.js#L1-L3


UncleBill:
Oops, it has been fixed in https://github.com/DimensionDev/Maskbook/commit/8f7a32694e95bccc7ad76af460ddcb24bb73152f.

Closing.

PR #5126, opened 2021-12-09
refactor: improve big number practice

septs:
closes #


packages/mask/src/plugins/Collectible/SNSAdaptor/ListingByHighestBidCard.tsx
```diff
@@ -79,7 +79,7 @@ export function ListingByHighestBidCard(props: ListingByHighestBidCardProps) {
             }),
             accountAddress: account,
             startAmount: Number.parseFloat(amount),
-            expirationTime: toUnixTimestamp(expirationDateTime),
+            expirationTime: getUnixTime(expirationDateTime),
```
guanbinrui, line 82: Are they equivalent?

packages/mask/src/plugins/ITO/SNSAdaptor/UnlockDialog.tsx
```diff
@@ -79,7 +78,7 @@ export function UnlockDialog(props: UnlockDialogProps) {
     //#endregion
     //#region amount
     const [rawAmount, setRawAmount] = useState('')
-    const amount = new BigNumber(rawAmount || '0').shiftedBy(token?.decimals ?? 0)
+    const amount = rightShift(rawAmount || '0', token?.decimals)
```
guanbinrui, line 81:
```suggestion
    const amount = rightShift(rawAmount || '0', token?.decimals ?? 0)
```

packages/mask/src/plugins/Collectible/SNSAdaptor/ListingByHighestBidCard.tsx
septs: https://github.com/DimensionDev/Maskbook/blob/32c223d898eb0e69891ad78147b9e20800372c84/packages/mask/src/plugins/Collectible/helpers.ts#L37-L39

image

yep, https://github.com/date-fns/date-fns/blob/master/src/getUnixTime/index.ts#L28

PR #5160, opened 2021-12-11
refactor: close a dialog from useRCD if its attached component will unmount

UncleBill:


UncleBill:
Can cause unintended results. Closing.

PR #5182, opened 2021-12-14
fix: should keep login status when restore persona which not have private key

NikoManiac:
1. Should keep login status when restore persona which not have private key
2. User can send NFT to self in dashboard


wenluomask:
This pr fixed https://mask.atlassian.net/browse/MASK-392

Checked.

wenluomask:
For the new commit, if I used ens in transfer, should be detected that it is myself.

wenluomask:
Checked.

guanbinrui (approved):

yanzhihong23 (approved):

PR #5294, opened 2021-12-25
feat: all i18n support hmr

Jack-Works:
## Description

Closes # (NO_ISSUE)

## Type of change

- [ ] Documentation

## Checklist

- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented on my code, particularly in hard-to-understand areas.
- [x] I have read [Internationalization Guide](https://github.com/DimensionDev/Maskbook/blob/develop/docs/i18n-guide.md) and moved text fields to the i18n JSON file.


packages/shared/src/UI/components/I18NextProviderHMR.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect, useState } from 'react'
+import { I18nextProvider, type I18nextProviderProps } from 'react-i18next'
+
+export function I18NextProviderHMR({ i18n, defaultNS, children }: React.PropsWithChildren<I18nextProviderProps>) {
```
UncleBill, line 4: I would suggest:

```ts
 export const I18NextProvider = process.env.NODE_ENV === 'development' ? I18NextProviderHMR : I18NextProviderWithoutHMR
```
Then you don't need to modify `<I18NextProvider />` in those node tree in above files. After all, they don't need to care about is the `<I18NextProvider />` hot-reload-able or not, do they?

PR #5303, opened 2021-12-26
feat: all i18n support hmr (#5294)

Jack-Works:
Let me try again to bring #5294 back


Jack-Works:
@jk234ert try again. It was a typo that breaks in the runtime

PR #5312, opened 2021-12-27
fix: walkaround of GeckoView's JSON serialization bug by stringifying the JSON object

jk234ert:
Due to GeckoView's bug

> GeckoView fails to serialize a JSON object with a nested array while communicating between JS and native, this makes the native side CANNOT receive some messages from `public-api`

This PR introduces a walkaround of stringifying the JSON before sending it to the message channel.
Tested positive on Android


Tlaster (approved):
LGTM!

Jack-Works:
Tracking: https://bugzilla.mozilla.org/show_bug.cgi?id=1681638

PR #5371, opened 2022-01-02
fix: i18n incorrect practice

septs:
https://www.i18next.com/translation-function/plurals

CC @Jack-Works


Tedko (approved):

Jack-Works:
in order to use this, we need to

- remove the locale kit we currently have in the CI
- support counting keys in @magic-works/i18n-codegen
- remove the current usage of `t("str")` and move to t.str()

septs:
@Jack-Works plz move this problem to jira.

PR #5167, opened 2021-12-13
feat: findtruman add property critical to votes

z384647062:
closes #

PR #5455, opened 2022-01-17
fix: no-explicit-any for plugin directory

septs:
https://github.com/DimensionDev/Maskbook/pull/5167#pullrequestreview-840159708

CleanShot 2022-01-17 at 09 40 46@2x

### DIFF
--- a/packages/.eslintrc.json
+++ b/packages/.eslintrc.json
@@ -83,7 +83,7 @@
     },
     "overrides": [
         {
-            "files": ["./plugins/**/*", "./mask/src/plugins/**/*"],
+            "files": ["./plugins/**/*"],
             "rules": {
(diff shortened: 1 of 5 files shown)


packages/.eslintrc.json
```diff
@@ -80,5 +80,13 @@
         "@typescript-eslint/prefer-string-starts-ends-with": "error",
         "@typescript-eslint/restrict-plus-operands": "off",
         "@typescript-eslint/restrict-template-expressions": "off"
-    }
+    },
+    "overrides": [
+        {
+            "files": ["./plugins/**/*"],
```
septs, line 86:
```
packages/mask/src/plugins/**/*
```

history problem too more (`✖ 72 problems (72 errors, 0 warnings)`)

#4609

PR #5464, opened 2022-01-17
fix: html entity

septs:


septs:
React only supported HTML 4 entity

UncleBill (approved):

nuanyang233 (approved):

PR #5467, opened 2022-01-17
chore: improve spellcheck

septs:


guanbinrui (dismissed):

cspell.json
```diff
@@ -21,6 +22,7 @@
         "AMPL",
         "Ampleforth",
         "anchorme",
+        "anomynous",
```
UncleBill: `anomynous` is a real typo for `anonymous`. You should you inline ignore instead.

guanbinrui (dismissed):

UncleBill:
@septs Let's lowercase the words in `cspell.json`

PR #5610, opened 2022-02-06
fix: add editorconfig checker

septs:


Jack-Works (approved):
interesting, but do we have so many files that do not managed by prettier?

guanbinrui (approved):

PR #5670, opened 2022-02-16
chore(eslint): ignore storybook-static in dashboard

UncleBill:
bundle files are minified, will take very long time and cause the task to fail.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -10,3 +10,4 @@ packages/netlify/storybook-static
 packages/netlify/sites
 public/
 scripts/
+packages/dashboard/storybook-static/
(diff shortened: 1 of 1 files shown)


septs (dismissed):

UncleBill:
for example https://github.com/DimensionDev/Maskbook/runs/5214436111?check_suite_focus=true

UncleBill:
for example https://github.com/DimensionDev/Maskbook/runs/5214436111?check_suite_focus=true

.eslintignore
```diff
@@ -10,3 +10,4 @@ packages/netlify/storybook-static
 packages/netlify/sites
 public/
 scripts/
+packages/dashboard/storybook-static/
```
Jack-Works, line 13: Also `packages/theme/storybook-static`?

UncleBill: done

Jack-Works (approved):

PR #5766, opened 2022-03-03
fix: check equality between two TypedMessage's

UncleBill:
## Description

Closes # (NO_ISSUE)

## Type of change

- [ ] Documentation

## Checklist

### DIFF
--- a/packages/typed-message/base/utils/index.ts
+++ b/packages/typed-message/base/utils/index.ts
@@ -1,13 +1,15 @@
-export * from './metadata'
 export * from './extract'
+export * from './metadata'
 
 import type {
-    TypedMessage,
-    SerializableTypedMessages,
-    SerializableTypedMessage,
     NonSerializableWithAltTypedMessage,
+    SerializableTypedMessage,
+    SerializableTypedMessages,
+    TypedMessage,
 } from '../base'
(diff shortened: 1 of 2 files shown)


packages/typed-message/base/utils/index.ts
```diff
@@ -26,6 +28,22 @@ export function isSerializableTypedMessage(x: TypedMessage): x is SerializableTy
 export function isTypedMessageEqual(message1: TypedMessage, message2: TypedMessage): boolean {
     if (message1.type !== message2.type) return false
     if (message1.meta !== message2.meta) return false
-    // perform a deep equal
-    return isEqual(message1, message2)
+    return renderMessage(message1) === renderMessage(message2)
```
UncleBill, line 31: This post https://twitter.com/Nate_FTXUS/status/1499219037552668674 , for example, flatten message is not equal to transformed message


https://github.com/DimensionDev/Maskbook/blob/801c3e8659b89ffad6d0648614153a88cc03daac/packages/mask/src/components/InjectedComponents/PostReplacer.tsx#L63

Jack-Works: hmm I think it should not equal. For example, `[[1, 2, 3]].flat()` is not the same as `[1, 2, 3]`

PR #5809, opened 2022-03-07
feat: rewrite shadow root css

Jack-Works:
close MF-97


packages/theme/src/ShadowRoot/Portal.tsx
```diff
@@ -41,64 +31,42 @@ export const NoEffectUsePortalShadowRootContext = createContext(false)
...
+        const container = shadow.appendChild(document.createElement('main'))
+        sheets.map((x) => x.addContainer(shadow))
+        return { container, root }
+    })
+    useEffect(
+        () => () => {
+            setTimeout(() => {
+                root.remove()
+                signal.current.abort()
+            }, 2000)
```
UncleBill, line 78: Missed cleanup function

packages/plugin-infra/src/utils/createInjectHooksRenderer.tsx
```diff
@@ -44,16 +45,20 @@ export function createInjectHooksRenderer<PluginDefinition extends Plugin.Shared
...
                 </PropsContext.Provider>
             ))
         return <>{all}</>
     }
     return memo(function PluginsInjectionHookRenderErrorBoundary(props: PropsType) {
         return (
-            <ErrorBoundary>
-                <PluginsInjectionHookRender {...props} />
-            </ErrorBoundary>
+            <span data-plugin-render="">
```
UncleBill, line 57: Is this `<span/>` necessary?

Don't see any css selector match `[data-plugin-render]` or some other usage

packages/plugin-infra/src/utils/createInjectHooksRenderer.tsx
```diff
@@ -44,16 +45,20 @@ export function createInjectHooksRenderer<PluginDefinition extends Plugin.Shared
             .filter(pickInjectorHook)
             .map((plugin) => (
                 <PropsContext.Provider key={plugin.ID} value={props}>
-                    <SinglePluginWithinErrorBoundary key={plugin.ID} plugin={plugin} />
+                    <NestedShadowRootIsolation data-plugin={plugin.ID}>
+                        <SinglePluginWithinErrorBoundary key={plugin.ID} plugin={plugin} />
```
UncleBill:
```suggestion
                        <SinglePluginWithinErrorBoundary plugin={plugin} />
```

guanbinrui (dismissed):

packages/theme/src/ShadowRoot/Portal.tsx
Jack-Works: it is in the effect clean up callback, no need to clean up again.

packages/plugin-infra/src/utils/createInjectHooksRenderer.tsx
Jack-Works: Not necessary. This span is used to wrap all plugins because there're many plugins and it's a noise. With this span, we can fold all plugins at once.

This `data-plugin-render` is not used anywhere. It's for a better view in the F12 DOM view.

guanbinrui (approved):

PR #6056, opened 2022-04-11
chore: fold single returns arrow function

septs:
https://eslint.org/docs/rules/arrow-body-style


UncleBill:
Please don't let it be so strict!
Personally, I would reject for this change.
**PLEASE KEEP IT STUPID AND SIMPLE!**

---

Besides, more code are been adding
image


--- 


Find real pain points and fix them. Code style "improvement" could be endless, investing so much time is not worth it.

There are some pain points I can see and feel:

## Icons

How to add a new icon: copy svg code from figma, then convert it into jsx manually.

https://github.com/DimensionDev/Maskbook/issues/3563

## Module/Package management

Many shared/utills package.

Dependencies are not well isolated. For example, plugins.

guanbinrui:
I like this. But this will make all of the existing pull requests into conflict.

Jack-Works:
I think no need for this rule. Both styles are easy to read, no need to limit how we write code if they're evenly good.

Issue #5770, opened 2022-03-03
[Demand] Enable Content Security Policy


Jack-Works:
https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

# Roadmap

- [x] Setup a CSP report server
- [ ] Have a CSP list and generate appropriate CSP policy string
- [ ] Enable `report-only` CSP in development mode
- [ ] Add good URLs to the allow list
- [ ] Enable `report-only` CSP in production mode
- [ ] Enable strict CSP in development mode
- [ ] Enable strict CSP in production mode

PR #6075, opened 2022-04-13
feat: enable csp report-only mode

Jack-Works:
part of #5770


Jack-Works:
blocked up upstream. I found there is no way to set a Report-Only CSP in WebExtension. Already reported to Chrome

PR #6129, opened 2022-04-22
fix(tips): some improvement

UncleBill:


packages/mask/src/plugins/Tips/SNSAdaptor/bodyViews/Wallets.tsx
```diff
@@ -4,7 +4,7 @@ import { Typography } from '@mui/material'
 import { memo, useCallback, useState } from 'react'
 import { useI18N } from '../../../../utils'
 import { DisconnectWalletDialog } from '../components/DisconnectDialog'
-import { WalletCom } from '../components/WalletCom'
+import { WalletItem } from '../components/WalletItem'
```
UncleBill, line 7: We rarely use `-Com` for `Component`

PR #5632, opened 2022-02-09
refactor: turn token picking into async call

UncleBill:
## UI Runtime

I introduce UI runtime in `packages/runtime/ui`.
I have a picture of building a runtime for both basic features and plugins, we might add more other runtimes to compose the final runtime. Each platform can implement its own runtime, and simply expose the API.

```tsx
<communicate>
  <runtime>
      <maskbook />
      <plugin1 />
      <plugin2 />
      <plugin3 />
  </runtime>
</communicate>
```

### runtime

```tsx
<chain-runtime>
  <ui-runtime>
    <plugin-runtime>{children}</plugin-runtime>
  </ui-runtime>
</chain-runtime>
```

### What would ui-runtime provide?

- system ui
  - toaster
- web3 ui toolkit
    - erc20picker
    - nft picker
    - gas settings
    - ??
- ???

## Usage example

```js
import { usePickToken } from '@masknet/ui-runtime'

const pickToken = usePickToken()
const handlePick = useCallback(() => {
  const picked = await pickToken({ disableNativeToken: true })
  setToken(picked)
}, [pickToken]) // should be added to dependency array.
```

https://mask.atlassian.net/browse/MF-313

PR #6181, opened 2022-04-28
refactor: replace TransactionDialog with useOpenShareTxDialog

UncleBill:
https://mask.atlassian.net/browse/MF-583
https://mask.atlassian.net/browse/MF-299
https://mask.atlassian.net/browse/MF-305
https://mask.atlassian.net/browse/MF-558


packages/mask/src/plugins/Gitcoin/SNSAdaptor/DonateDialog.tsx
```diff
@@ -101,49 +99,37 @@ export function DonateDialog(props: DonateDialogProps) {
     const amount = rightShift(rawAmount || '0', token?.decimals)
     // #endregion
 
-    // #region blocking
-    const [donateState, donateCallback, resetDonateCallback] = useDonateCallback(address ?? '', amount.toFixed(), token)
-    // #endregion
-
-    // #region transaction dialog
+    const [loading, donateCallback] = useDonateCallback(address ?? '', amount.toFixed(), token)
```
guanbinrui: Why return the loading state here? The `donateCallback` will return a `AsyncResult`.

UncleBill:
@DimensionDev/qa Please help to verify.

packages/mask/src/plugins/Gitcoin/SNSAdaptor/DonateDialog.tsx
UncleBill: Fixed.

packages/shared/src/hooks/useOpenShareTxDialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+    },
+}))
+
+interface ShareTransactionOptions {
+    title?: string
+    message?: string
+    content?: string
+    hash: string
+    buttonLabel?: string
+    // TODO Expose onShare until we have share API added our runtime
```
Jack-Works, line 40: Can you explain more clearly?

Jack-Works (changes requested):

packages/shared/src/hooks/useOpenShareTxDialog.tsx
UncleBill: We can't access `activatedSocialNetworkUI` outside `packages/mask` now,  although it should be a feature of a runtime for plugin.

I once introduced `runtime` at https://github.com/DimensionDev/Maskbook/pull/5632#issue-1128316587 .

```
<chain-runtime>
  <ui-runtime>
    <plugin-runtime>{children}</plugin-runtime>
  </ui-runtime>
</chain-runtime>
```

We can separate different concepts into individual layers, and we can implement each of them across platforms. iOS, android, and we might have desktop in the future, and even for mocking/testing.

If we implement mocking runtime, we can easily test with headless JS runtime. For example, we can turn UI interaction into function-call.

```
const share = () => Promise.resolve(true)
const pickToken = () => Promise.resolve({name: 'Mask'})
const web3Provider = () => connnectToLocalNetwork()
```

Jack-Works: You may be interested to add shareTo at `Plugin.SNSAdaptor.SNSAdaptorContext` at packages/plugin-infra/src/types.ts.

packages/shared/src/contexts/common/Confirm/Dialog.tsx
```diff
@@ -0,0 +1,39 @@
+import { makeStyles } from '@masknet/theme'
+import { Button, DialogActions, DialogContent } from '@mui/material'
+import type { FC } from 'react'
+import { useSharedI18N } from '../../../locales'
+import { InjectedDialog, InjectedDialogProps } from '../../components'
+import type { ConfirmOptions } from './types'
+
+const useStyles = makeStyles()((theme) => ({
```
Jack-Works, line 8: Is this folder duplicated with ../ShareDialog? Please check.

packages/shared/src/hooks/useOpenShareTxDialog.tsx
Jack-Works: ping

Jack-Works:
Most of this PR LGTM, but still 2 problems

packages/shared/src/contexts/common/Confirm/Dialog.tsx
UncleBill: Indeed, has removed.

packages/shared/src/hooks/useOpenShareTxDialog.tsx
UncleBill: I would like to do it within another patch.

Jack-Works: It's ok, please open an issue so we won't forget to do it in the future.

UncleBill: > It's ok, please open an issue so we won't forget to do it in the future.

https://github.com/DimensionDev/Maskbook/issues/6233

packages/shared/src/contexts/UITaskManager.tsx
```diff
@@ -0,0 +1,99 @@
+import { defer, DeferTuple } from '@dimensiondev/kit'
+import { EMPTY_LIST } from '@masknet/shared-base'
+import { createContext, createElement, FC, FunctionComponent, PropsWithChildren, useMemo, useState } from 'react'
+import type { InjectedDialogProps } from './components'
+
+type ExtendPromise<T> = Promise<T> & {
```
Jack-Works: Can you explain the use case of this?

UncleBill: Intend to provide a way to close a dialog programmably.
There is no use case right now indeed.

Jack-Works: Is it possible to use things like `{promise: Promise<T>, close?: () => void}`?

UncleBill: > Is it possible to use things like `{promise: Promise<T>, close?: () => void}`?

Then the common usage would be wordy

```ts
const openDialog = useOpenDialog()
await openDialog(options).promise
```

----

> Intend to provide a way to close a dialog programmably.

Use case for `close()`

```ts
const openDialog = useOpenDialog()
const promise = openDialog(options)
const timer = setTimeout(promise.close, 1e5)
await promise
clearTimeout(timer)
```

Jack-Works: OK, this is interesting. And where are you attaching the `close` method?

UncleBill:
```ts
                    const extendedPromise = Object.assign(promise, {
                        close() {
                            resolve(null)
                            return promise
                        },
                    })
                    const newTask: Task = { id, promise, resolve, reject, taskOptions: options }
                    setTasks((list) => [...list, newTask])
                    promise.then(() => {
                        removeTask(id)
                    })
                    return extendedPromise
```

packages/shared/src/contexts/UITaskManager.tsx
```diff
@@ -0,0 +1,99 @@
...
+                setTasks((list) => list.filter((t) => t.id !== id))
+            }
+
+            return {
+                show(options: TaskOptions) {
+                    const [promise, resolve, reject] = defer<Result | null>()
+                    id += 1
+                    const extendedPromise = Object.assign(promise, {
+                        close() {
+                            resolve(null)
```
Jack-Works: You should use reject instead of resolve null.

packages/shared/src/contexts/UITaskManager.tsx
```diff
@@ -0,0 +1,99 @@
...
+    taskResolveField: ResolveProp,
+) => {
+    type ContextOptions = ContextOptionsGeneric<TaskOptions, Result | null>
+    const TaskManagerContext = createContext<ContextOptions>(null!)
+    let id = 0
+
+    type TaskDeferTuple = DeferTuple<Result | null>
+    interface TaskBase {
+        id: number
+        promise: Promise<Result | null>
```
Jack-Works, line 31:
```suggestion
        promise: Promise<Result>
```

packages/shared/src/contexts/UITaskManager.tsx
```diff
@@ -0,0 +1,99 @@
...
+    ResolveProp extends CallableKeys<Props> & CallableKeys<TaskOptions>,
+>(
+    Component: ComponentType<Props>,
+    taskResolveField: ResolveProp,
+) => {
+    type ContextOptions = ContextOptionsGeneric<TaskOptions, Result | null>
+    const TaskManagerContext = createContext<ContextOptions>(null!)
+    let id = 0
+
+    type TaskDeferTuple = DeferTuple<Result | null>
```
Jack-Works, line 28:
```suggestion
    type TaskDeferTuple = DeferTuple<Result>
```

packages/shared/src/contexts/UITaskManager.tsx
```diff
@@ -0,0 +1,99 @@
...
+        const [tasks, setTasks] = useState<Task[]>(EMPTY_LIST)
+
+        const contextValue = useMemo(() => {
+            const removeTask = (id: number) => {
+                setTasks((list) => list.filter((t) => t.id !== id))
+            }
+
+            return {
+                show(options: TaskOptions) {
+                    const [promise, resolve, reject] = defer<Result | null>()
```
Jack-Works, line 47:
```suggestion
                    const [promise, resolve, reject] = defer<Result>()
```

Jack-Works (changes requested):

packages/shared/src/contexts/UITaskManager.tsx
```diff
@@ -0,0 +1,99 @@
...
+ */
+export const createUITaskManager = <
+    TaskOptions extends Record<string, any>,
+    Result,
+    Props extends BaseDialogProps,
+    ResolveProp extends CallableKeys<Props> & CallableKeys<TaskOptions>,
+>(
+    Component: ComponentType<Props>,
+    taskResolveField: ResolveProp,
+) => {
```
Jack-Works: I suggest the following changes:

- Remove `ExtendPromise`, and change to use `AbortController`.
- Remove `ResolveProp` type, and enforce the component to use a blessed name, e.g. `onSubmit(value: T): void`.

Therefore, I suggest changing the type signature to the below:

```ts
export function createUITaskManager<
    Opts extends { open: boolean, onClose(): void, onSubmit(value: T): void },
    T
>(component: React.ComponentType<Opts>): [
    useUI: () => (options: Options<Opts>, signal?: AbortSignal) => Promise<T>,
    provider: React.ComponentType<React.PropsWithChildren<{}>>
]
```

Usage:

```ts
const showDialog = useUI()

async function onClick() {
    const abort = new AbortController()
    const result = await showDialog({ title: 'abc' }, abort.signal)
}
```

Jack-Works (changes requested):

packages/shared/src/contexts/UITaskManager.tsx
UncleBill: No, aborting should not raise an exception.
```ts
const pickeToken = await pickToken()
// get null by just closing the dialog
// like `window.prompt`
const text = prompt('what's your name?')
```

Jack-Works: Then what if `Result` contains `null`? You cannot distinguish between these two different states because you express them both in `null`.

UncleBill: >Then what if Result contains null?

That means "I don't care about the result at all"

Jack-Works: You should distinguish result/close with one of the following way:

- Promise rejection
- `Some<T>`
- Tagged union (`{ type: 'close' } | { type: 'resolve', value: T }`)

UncleBill: I prefer to keep it simple.
We still have chance to do such refactor if we need it.

packages/shared/src/contexts/common/Confirm/types.ts
```diff
@@ -0,0 +1,8 @@
+import type { ReactNode } from 'react'
+
+export interface ConfirmOptions {
+    title?: string
+    content: ReactNode | string
+    confirmLabel?: string
+    onSubmit?(result: boolean | null): void
```
Jack-Works, line 7: In this case, I suggest using treat `close dialog` as `cancel`, which means they should both be represented by `false`.

packages/shared/src/hooks/useOpenShareTxDialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+    )
+})
+
+export function useOpenShareTxDialog() {
+    const showConfirm = useShowConfirm()
+    const t = useSharedI18N()
+
+    return async ({ title, message, content, hash, buttonLabel, onShare }: ShareTransactionOptions) => {
+        let shared = false
+        await showConfirm({
```
Jack-Works: Isn't the result come back from showConfirm? 🤔 Why you need an extra variable `shared` to record the shared state?

packages/shared/src/contexts/UITaskManager.tsx
```diff
@@ -0,0 +1,89 @@
+import { defer, DeferTuple } from '@dimensiondev/kit'
+import { EMPTY_LIST } from '@masknet/shared-base'
+import { createContext, createElement, FC, ComponentType, PropsWithChildren, useMemo, useState } from 'react'
+import type { InjectedDialogProps } from './components'
+
+interface ContextOptionsGeneric<T, R> {
```
Jack-Works:
```suggestion
interface ContextOptions<T, R> {
```

packages/shared/src/contexts/common/Confirm/types.ts
UncleBill: Closing the dialog means "no data input", then `null` makes sense.

In general, a `confirm` dialog should provide a cancel button for user input, but no close button.


image


And we do have such design, but it has not yet been implemented.
We will implement it as the final `Confirm` dialog.

image

lelenei (approved):

nuanyang233 (approved):

guanbinrui:
Replace `useCallback` into `useAsyncFn`.

UncleBill:
> Replace `useCallback` into `useAsyncFn`.

Already done.

Jack-Works (approved):
Although I still think we should use reject instead of null to represent the cancel state.

packages/dashboard/src/pages/Wallets/components/Transfer/TransferERC721.tsx
```diff
@@ -204,23 +202,19 @@ export const TransferERC721 = memo(() => {
...
-                await transferCallback(data.tokenId, registeredAddress, gasConfig)
+                hash = await transferCallback(data.tokenId, registeredAddress, gasConfig)
+            }
+            if (typeof hash === 'string') {
+                navigate(DashboardRoutes.WalletsHistory)
             }
-            return
         },
-        [transferCallback, contract?.address, gasConfig, registeredAddress, Utils?.isValidDomain],
+        [transferCallback, contract?.address, gasConfig, registeredAddress, Utils?.isValidDomain, navigate],
```
guanbinrui, line 217: No navigate in deps.

packages/mask/src/plugins/ArtBlocks/SNSAdaptor/PurchaseDialog.tsx
```diff
@@ -94,36 +89,27 @@ export function PurchaseDialog(props: ActionBarProps) {
...
+            const hash = await purchaseCallback()
+            if (typeof hash !== 'string') return
+            await openShareTxDialog({
+                hash,
+                onShare() {
+                    activatedSocialNetworkUI.utils.share?.(shareText)
+                },
+            })
+            onClose()
+        } catch (err: any) {
```
guanbinrui, line 105: Avoid `any`.

PR #6190, opened 2022-04-30
chore: improve eslint rules

septs:


packages/.eslintrc.json
```diff
@@ -102,7 +104,15 @@
...
-                "@typescript-eslint/no-explicit-any": "error"
+                "@typescript-eslint/no-explicit-any": "error",
+                "@dimensiondev/no-number-constructor": "error",
+                "@dimensiondev/type/no-empty-literal": [
+                    "off",
+                    {
+                        "array": "Use `EMPTY_LIST` instead of `[]`",
+                        "object": "Use `EMPTY_OBJECT` instead of `{}`"
+                    }
+                ]
```
septs, line 115: plz design a file match pattern on the eslint rule

septs: @guanbinrui @Jack-Works

packages/mask/background/database/persona/consistency.ts
```diff
@@ -127,6 +127,7 @@ async function* checkProfileLink(
     }
 }
 
+// eslint-disable-next-line @dimensiondev/type/no-const-enum
```
Jack-Works: I think if the const enum is not exported, it's ok to use. The problem of const enum is that it does not support cross-file transpile, and const enum not exported does not have this problem.

packages/web3-providers/src/twitter/index.ts
```diff
@@ -21,6 +21,11 @@ function getScriptContentMatched(content: string, regexp: RegExp) {
 }
 
 function getCSRFToken() {
+    // use `cookieStore` replace the api call?
+    // see https://developer.mozilla.org/en-US/docs/Web/API/CookieStore
```
Jack-Works: Need polyfill on Safari and Firefox 🤔

Jack-Works (dismissed):

packages/.eslintrc.json
Jack-Works: have no idea

Jack-Works (approved):

packages/.eslintrc.json
UncleBill: Both `EMPTY_LIST` and `EMPTY_OBJECT` were introduce to help to reduce re-rendering. It's hard to find out if you need or not by doing code analysis. (if only linter can understand react 😂 )

I think we should guide by docs rather than linter.

UncleBill: I think we can close this, and start to write some `best practice`.

PR #6192, opened 2022-05-01
chore: upgrade to pnpm7

Jack-Works:
🤔

### DIFF
--- a/.lintstagedrc.mjs
+++ b/.lintstagedrc.mjs
@@ -2,6 +2,6 @@ export default {
     '*': 'prettier --write --ignore-unknown',
     'packages/**/*.{ts,tsx,js,jsx}': 'eslint -c packages/.eslintrc.json --fix',
     'packages/mask/**/*': () => 'gulp locale-kit --sync-keys --remove-unused-keys',
-    'packages/web3-constants/**/*': 'pnpm start --filter web3-constants',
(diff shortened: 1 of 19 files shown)


.lintstagedrc.mjs
```diff
@@ -2,6 +2,6 @@ export default {
     '*': 'prettier --write --ignore-unknown',
     'packages/**/*.{ts,tsx,js,jsx}': 'eslint -c packages/.eslintrc.json --fix',
     'packages/mask/**/*': () => 'gulp locale-kit --sync-keys --remove-unused-keys',
-    'packages/web3-constants/**/*': 'pnpm start --filter web3-constants',
+    'packages/web3-constants/**/*': 'pnpm --filter ./packages/web3-constants start',
```
septs:
```suggestion
    'packages/web3-constants/**/*': () => 'pnpm --filter ./packages/web3-constants start',
```

reduce file list input on the command

Jack-Works: Sorry, but what's the difference?

septs: lint-staged default inject file list into command arguments list tail

https://github.com/okonet/lint-staged#example-run-tsc-on-changes-to-typescript-files-but-do-not-pass-any-filename-arguments

PR #6217, opened 2022-05-09
refactor: update promote text for plugins.

UncleBill:
https://mask.atlassian.net/browse/MF-468


packages/scripts/src/locale-kit/ast.ts
```diff
@@ -1,5 +1,7 @@
 import ts from 'typescript'
 
+const transformIdentifiers = ['tr', 't']
```
Jack-Works: Move out to its own PR?

packages/.eslintrc.json
```diff
@@ -77,7 +77,7 @@
         "@dimensiondev/string/no-unneeded-to-string": "error",
         "@dimensiondev/type/no-instanceof-wrapper": "error",
         "@dimensiondev/type/no-wrapper-type-reference": "error",
-        "@dimensiondev/unicode/specific-set": "error",
+        "@dimensiondev/unicode/specific-set": "off",
```
septs: why?

septs: this is security rule

packages/mask/src/plugins/RedPacket/SNSAdaptor/index.tsx
```diff
@@ -94,7 +94,7 @@ const sns: Plugin.SNSAdaptor.Definition = {
     ApplicationEntries: [
         (() => {
             const icon = <RedPacketIcon />
-            const name = <Trans i18nKey="plugin_red_packet_name" />
+            const name = <Trans i18nKey="com.maskbook.red_packet:name" />
```
Jack-Works: Wait... You should use 

```js
import { useI18N } from '../locales'
const t = useI18N()
t.name()
```

Jack-Works: use i18n hooks.

packages/mask/src/plugins/RedPacket/SNSAdaptor/RedPacketInHistoryList.tsx
```diff
@@ -332,7 +329,7 @@ export function RedPacketInHistoryList(props: RedPacketInHistoryListProps) {
                         </Typography>
                         <Typography variant="body1" className={classes.footerInfo}>
                             <Trans
```
Jack-Works: use TypedTrans from redpacket i18n folder.

packages/scripts/src/locale-kit/ast.ts
Jack-Works: never mind. you can keep it  in this pr.

packages/.eslintrc.json
UncleBill: A trade-off between security and convenience.
Converting emoji into unicode make them hard to read and modify. We should look for another solution. I think scanning for zero-width characters is enough.

packages/mask/src/plugins/RedPacket/SNSAdaptor/index.tsx
UncleBill: Not all `<Trans />` here are in React component directly. Using `<Trans />` directly is more convenient.

packages/mask/src/plugins/RedPacket/SNSAdaptor/RedPacketInHistoryList.tsx
UncleBill: done.

UncleBill (commented):
I think a draft PR is not supposed to be reviewed except you get a request.

packages/.eslintrc.json
Jack-Works: Are emojis very common? If not, you can use eslint-disable

packages/mask/src/components/CompositionDialog/useSubmit.ts
```diff
@@ -4,14 +4,17 @@ import { RedPacketMetadataReader } from '../../plugins/RedPacket/SNSAdaptor/help
...
 import type { SubmitComposition } from './CompositionUI'
 import { useLastRecognizedIdentity } from '../DataSource/useActivatedUI'
 import { isFacebook } from '../../social-network-adaptor/facebook.com/base'
+import { ITO_MetadataReader } from '../../plugins/ITO/SNSAdaptor/helpers'
+import { FileInfoMetadataReader } from '@masknet/plugin-file-service'
+import type { Meta } from '@masknet/typed-message'
 
 export function useSubmit(onClose: () => void, reason: 'timeline' | 'popup' | 'reply') {
-    const { t } = useI18N()
+    const { t: t } = useI18N()
```
Jack-Works:
```suggestion
    const { t } = useI18N()
```

packages/mask/src/plugins/Gitcoin/SNSAdaptor/index.tsx
```diff
@@ -42,8 +42,8 @@ const sns: Plugin.SNSAdaptor.Definition = {
         {
             ApplicationEntryID: base.ID,
             category: 'dapp',
-            description: <Trans i18nKey="plugin_gitcoin_description" />,
-            name: <Trans i18nKey="plugin_gitcoin_name" />,
+            description: <Trans ns={PluginId.Gitcoin} i18nKey="description" />,
```
Jack-Works, line 41: Those two fields accepts I18NField. Please use i18nField (`{ i18nKey: '__description', fallback: '...' }`)

packages/mask/src/plugins/RedPacket/SNSAdaptor/index.tsx
```diff
@@ -94,7 +94,7 @@ const sns: Plugin.SNSAdaptor.Definition = {
     ApplicationEntries: [
         (() => {
             const icon = <RedPacketIcon />
-            const name = <Trans i18nKey="plugin_red_packet_name" />
+            const name = <Trans ns={PluginId.RedPacket} i18nKey="name" />
```
Jack-Works, line 97: This field support I18NField.

packages/mask/src/plugins/RedPacket/SNSAdaptor/index.tsx
```diff
@@ -118,7 +118,7 @@ const sns: Plugin.SNSAdaptor.Definition = {
                 appBoardSortingDefaultPriority: 1,
                 marketListSortingPriority: 1,
                 icon,
-                description: <Trans i18nKey="plugin_red_packet_description" />,
+                description: <Trans ns={PluginId.RedPacket} i18nKey="description" />,
```
Jack-Works, line 128: This field support I18NField.

packages/mask/src/components/CompositionDialog/useSubmit.ts
```diff
@@ -75,3 +65,34 @@ async function pasteImage(
         reason,
     })
 }
+
+// TODO: Provide API to plugin to postprocess post content,
+// then we can move these -PreText's and meta readrs into plugin's own context
+function decorateEncryptedText(encrypted: string, t: I18NFunction, meta?: Meta) {
+    const hasOfficalAccount = isTwitter(activatedSocialNetworkUI) || isFacebook(activatedSocialNetworkUI)
+    const officalAccount = isTwitter(activatedSocialNetworkUI) ? t('twitter_account') : t('facebook_account')
```
Jack-Works: Maybe move these two lines into a function?

packages/plugins/FileService/src/index.ts
```diff
@@ -1,6 +1,8 @@
 import { registerPlugin } from '@masknet/plugin-infra'
 import { base } from './base'
 
+export * from './helpers'
```
Jack-Works: Only export the util you want to expose.

packages/mask/src/plugins/RedPacket/SNSAdaptor/index.tsx
UncleBill: ReactNode one looks cleaner, doesn't it?
```diff
-            description: <Trans ns={PluginId.Gitcoin} i18nKey="description" />,
+            description: { i18nKey: `${PluginId.Gitcoin}:description`, fallback: 'description' },
```

In addition, all `ApplicationEntries` for Definition of all plugins are using `<Trans />` instead of `I18NField`, I think it's better to keep them align.

Jack-Works: If you're not using them by `t.description` or `Translate.description`, please make sure they start with `__`, in this case, `__description`.

We'll have a locale kit in the future to remove unused fields based on the type analysis, starting from `__` is a escape path, "don't remove this field even it is not used in the type level"

packages/.eslintrc.json
Jack-Works: ping

UncleBill: Not only emojis, but also some other symbols like Ξ ₿.

packages/mask/src/components/CompositionDialog/useSubmit.ts
UncleBill: Need to pass down `t`, that will be complicated.

packages/mask/src/plugins/Collectible/SNSAdaptor/MakeOfferDialog.tsx
```diff
@@ -111,7 +112,8 @@ export function MakeOfferDialog(props: MakeOfferDialogProps) {
             expirationTime: !isAuction ? getUnixTime(expirationDateTime) : undefined,
             paymentTokenAddress: token.value.type === EthereumTokenType.Native ? undefined : token.value.address,
         })
-    }, [asset?.value, token, account, amount, expirationDateTime, isAuction])
+        activatedSocialNetworkUI.utils.share?.(t('promote_collectible'))
+    }, [asset?.value, token, account, amount, expirationDateTime, isAuction, t])
```
septs, line 123:
```suggestion
    }, [asset?.value, token, account, amount, expirationDateTime, isAuction])
```

packages/.eslintrc.json
Jack-Works: cc @septs can we have an allow list for this rule? It will be convenient.

packages/mask/src/plugins/Collectible/SNSAdaptor/MakeOfferDialog.tsx
UncleBill: `t` is used. `t('promote_collectible')`

Jack-Works: IIRC `t` will change if the language changes.

UncleBill:
Don't merge until #6181 gets merge into develop branch.

packages/mask/src/plugins/RedPacket/SNSAdaptor/index.tsx
Jack-Works: ping

packages/.eslintrc.json
Jack-Works: ping @septs

packages/mask/src/plugins/RedPacket/SNSAdaptor/index.tsx
UncleBill: Instead of `__` prefix, can we define some reserved keys, such as `name`, `description`, etc, which can't be removed even they are not used, as each plugin needs to reserve them.


Our i18n keys are all in `snake_case`, we can force the reserved keys in `UpperCamelCase`, and normal keys in `lowerCamelCase` or `snake_case`.


Update: locale-kit has been removed, we can revisit it when we want it back.

Jack-Works: yes we can make some name blessed, but we still need a mark if devs want to use their own reserved names. `__` has already baked into the plugin infra

UncleBill: TBH, `__` looks weird. 😂

Jack-Works: Do you have a better suggestion? We can use another prefix/suffix.

UncleBill: My suggestion is  using `UpperCamelCase` for reserved keys. (Or just capitalize the first letter)

packages/mask/src/plugins/Gitcoin/SNSAdaptor/index.tsx
UncleBill: `description` for Gitcoin plugin is so long, it will mess up the code.

```json
    "description": "Display specific information of Gitcoin projects, donate to a project directly on social media.",
```

```diff
-           description: <Trans ns={PluginId.Gitcoin} i18nKey="description" />,
+           description: {
+               i18nKey: 'description',
+               fallback:
+                   'Display specific information of Gitcoin projects, donate to a project directly on social media.',
+           },
```

BTW, the `fallback` can easily outdate without being noticed. And managing text/translation in too many places looks like not following the DRY principle.

Jack-Works: Agree. Track this issue on #6357.

packages/mask/src/plugins/RedPacket/SNSAdaptor/index.tsx
Jack-Works: Track this issue on #6357.

packages/.eslintrc.json
Jack-Works: This rule supports allow list. https://dimensiondev.github.io/eslint-plugin/src/rules/unicode/specific-set

Please allow characters you need to use here.

Note: For the code review perspective, `₿` should not be allowed because it looks visually like `B`

Jack-Works: I have no other change requests on this PR. Once this one fixed, I will approve it

Jack-Works (approved):

guanbinrui (approved):

PR #6235, opened 2022-05-12
fix(gas): don't arbitrarily fallback gas to zero if fails to estimate gas fee

UncleBill:
https://mask.atlassian.net/browse/MF-582


Jack-Works (approved):

UncleBill:
Closes since outdated.

PR #6238, opened 2022-05-12
fix: lint disable

septs:


packages/mask/src/extension/background-script/EthereumServices/providers/MetaMask.ts
```diff
@@ -50,8 +50,7 @@ async function onMessage(message: {
 }
 
 export async function createProvider() {
-    // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
-    if (provider && provider.chainId !== null) return provider
+    if (!provider?.chainId) return provider
```
guanbinrui:
```suggestion
    if (provider?.chainId) return provider
```

guanbinrui (dismissed):

packages/dashboard/src/pages/SignUp/steps/PersonaRecovery.tsx
```diff
@@ -19,35 +19,39 @@ export const PersonaRecovery = () => {
...
+                }
 
                 await changeCurrentPersona(identifier)
                 showSnackbar(t.create_account_persona_successfully(), { variant: 'success' })
 
                 await delay(300)
                 navigate(`${DashboardRoutes.SignUp}/${SignUpRoutePath.ConnectSocialMedia}`)
             } catch (error) {
-                setError((error as Error).message)
+                if (error instanceof Error) {
```
Jack-Works: This change is not identical to the original one. If an error is not an Error instance, before, it will call `setError` but now it will do nothing.

Jack-Works (dismissed):

packages/mask/background/network/gun/encryption/queryPostKey.ts
```diff
@@ -177,33 +177,13 @@ namespace Version38Or39 {
...
-            }
-            return window.btoa(binary)
-        }
+    async function work(data: Uint8Array | string, salt: Uint8Array | string) {
+        if (typeof data === 'string') data = new TextEncoder().encode(data)
+        if (typeof salt === 'string') salt = new TextEncoder().encode(salt)
+        const key = await crypto.subtle.importKey('raw', data, { name: 'PBKDF2' }, false, ['deriveBits'])
+        const params: Pbkdf2Params = { name: 'PBKDF2', iterations: 100000, salt, hash: { name: 'SHA-256' } }
+        const derived = await crypto.subtle.deriveBits(params, key, 512)
+        return window.btoa(String.fromCharCode(...new Uint8Array(derived)))
```
Jack-Works, line 186: About this line, I don't know if `btoa+fromCharCode` is equivalent to the original algorithm

septs: CleanShot 2022-05-16 at 11 37 25@2x

Jack-Works (approved):

Lanttcat (approved):

PR #6246, opened 2022-05-13
refactor: replace use-subscription

SukkaW:
## Description

Closes # (NO_ISSUE)

## Type of change

- [ ] Documentation

## Checklist

- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
  - [x] I have removed all in development `console.log`s
  - [x] I have removed all commented code.
- [x] I have commented on my code, particularly in hard-to-understand areas.

------

Back in 2019, React released the first version of `use-subscription` (https://github.com/facebook/react/pull/15022). At the time, we only have very limited information about concurrent rendering.

In 2020, React provides a first-party official API `useMutableSource` (https://github.com/reactjs/rfcs/pull/147, https://github.com/facebook/react/pull/18000):

> ... enables React components to safely and efficiently read from a mutable external source in Concurrent Mode.

React 18 introduces `useMutableSource`'s replacement `useSyncExternalStore` (see details here: https://github.com/reactwg/react-18/discussions/86), and React changes `use-subscription` implementation to use `useSyncExternalStore` directly: https://github.com/facebook/react/pull/24289

> In React 18, `React.useSyncExternalStore` is a built-in replacement for `useSubscription`.
> 
> This PR makes `useSubscription` simply use `React.useSyncExternalStore` when available. For pre-18, it uses a `use-sync-external-store` shim which is very similar in `use-subscription` but fixes some flaws with concurrent rendering.

And according to `use-subscription`:

> You may now migrate to [`use-sync-external-store`](https://www.npmjs.com/package/use-sync-external-store) directly instead, which has the same API as `React.useSyncExternalStore`. The `use-subscription` package is now a thin wrapper over `use-sync-external-store` and will not be updated further.

The PR does exactly that:

- Removes `use-subscription`
- Adds the `use-sync-external-store` to the dependencies.
  - The `use-sync-external-store` package enables compatibility with React 16 and React 17.
- Replace `useSubscription` usage with `useSyncExternalStoreWithSelector`
  - Unlike `useSyncExternalStore` shim, `useSyncExternalStoreWithSelector` provides the built-in memo support to prevent infinite loop rendering.


packages/shared-base/src/utils/subscription.ts
```diff
@@ -1,8 +1,21 @@
...
-import type { Subscription } from 'use-subscription'
 import { None, Option, Some } from 'ts-results'
 
+type Unsubscribe = () => void
+
+export interface Subscription<T> {
+    /**
+     * (Synchronously) returns the current value of our subscription.
+     */
+    getCurrentValue: () => T
```
Jack-Works, line 11: I think you should add a `getServerValue?(): T` to `Subscription` type, and export the following hooks from the `shared` (React hooks should not inside the `shared-base`) package:

```js
function useSubscription<T>(sub: Subscription<T>): T {
    return useSyncExternalStoreWithSelector(
        sub.subscribe,
        sub.getCurrentValue,
        sub.getServerValue || sub.getCurrentValue,
        (s) => s,
    )
}
```

It will make the migration easier.

Jack-Works (changes requested):
Thanks for your interest! I have some review suggestions

packages/shared-base/src/utils/subscription.ts
SukkaW: The approach is only good for migration, not for future refactoring.

Consider this: What happened if different logic for `getServerValue` would be applied across different `useSyncExternalStoreWithSelector` usage in the future? What happened if multiple subscriptions will be merged into one store and then fully utilize the `selector` feature in the future?

IMHO, we should keep `useSyncExternalStoreWithSelector` as it is.

Jack-Works: We haven't made a plan to use a full-featured store like zustand or redux, thus I think easy migration is more important.

> What happened if different logic for getServerValue would be applied across different useSyncExternalStoreWithSelector usage in the future?

Can you explain this part?

PR #6264, opened 2022-05-16
chore(locale-kit): scan for plural/context/nested i18n keys

### LINKED ISSUE #6221: [Demand] locale-kit should not remove nesting keys
Nesting is a useful feature of `i18next`:  https://www.i18next.com/translation-function/nesting

image

Update the script to support nesting or disable it from commit hook.

UncleBill:
Set both `pluralSeparator` and `contextSeparator` to `$`

closes #6221

### DIFF
--- a/packages/shared-base/src/i18n/instance.ts
+++ b/packages/shared-base/src/i18n/instance.ts
@@ -15,6 +15,7 @@ if (!i18n.isInitialized) {
     i18n.use(Detector).init({
         keySeparator: false,
         interpolation: { escapeValue: false },
+        contextSeparator: '$',
         fallbackLng: {
             'zh-CN': ['zh-TW', 'en'],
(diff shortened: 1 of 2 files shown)


packages/shared-base/src/i18n/instance.ts
```diff
@@ -15,6 +15,8 @@ if (!i18n.isInitialized) {
     i18n.use(Detector).init({
         keySeparator: false,
         interpolation: { escapeValue: false },
+        pluralSeparator: '$',
+        contextSeparator: '$',
```
Jack-Works: how will it affect the use of plural and context?

UncleBill: Existed code needs update.

I see only one use case, get it fixed.

PR #6274, opened 2022-05-17
chore: no-useless-{catch,rename}

septs:


guanbinrui (approved):

packages/mask/shared-ui/locales/en-US.json
```diff
@@ -65,7 +65,6 @@
     "compose_encrypt_visible_to_private": "Private",
     "compose_encrypt_visible_to_private_sub": "Just Me",
     "compose_shared_friends": "1 friend",
-    "compose_shared_friends_other": "{{count}} friends",
```
Jack-Works, line 68: @septs this is wrong

septs: suggest, disable locale-kit

PR #6317, opened 2022-05-20
refactor: move conditional text logic into i18n layer

UncleBill:
Set `contextSeparator` to `$`

https://mask.atlassian.net/browse/MF-773

- [ ] Would rewrite some usage which as following screenshot shows

image

### DIFF
--- a/packages/mask/src/utils/i18n-next-ui.ts
+++ b/packages/mask/src/utils/i18n-next-ui.ts
@@ -3,11 +3,17 @@ import type en from '../../shared-ui/locales/en-US.json'
 import type { i18NextInstance, TranslateOptions } from '@masknet/shared-base'
 import { SupportedLanguages } from '@masknet/public-api'
 
-export type I18NFunction = <TKeys extends keyof typeof en>(
+type PluralsSuffix = 'zero' | 'one' | 'two' | 'few' | 'many' | 'other'
+
+type LocaleKeys = keyof typeof en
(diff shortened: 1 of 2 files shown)


packages/mask/src/utils/i18n-next-ui.ts
```diff
@@ -3,11 +3,17 @@ import type en from '../../shared-ui/locales/en-US.json'
...
 import { SupportedLanguages } from '@masknet/public-api'
 
-export type I18NFunction = <TKeys extends keyof typeof en>(
+type PluralsSuffix = 'zero' | 'one' | 'two' | 'few' | 'many' | 'other'
+
+type LocaleKeys = keyof typeof en
+type ExtendBaseKeys<K> = K extends `${infer B}$${string}` ? B | K : K extends `${infer B}_${PluralsSuffix}` ? B | K : K
+type AvailableLocaleKeys = ExtendBaseKeys<LocaleKeys>
+
+export type I18NFunction = <TKeys extends AvailableLocaleKeys>(
```
Jack-Works, line 12: I suggest to stop using this, use i18n exported from `packages/mask/shared-ui/locales`, it provides context support.

UncleBill:
Keep closed until I start to work on.

packages/mask/src/utils/i18n-next-ui.ts
UncleBill: There are so many places. It's no easy to migrate in a single patch.

PR #6338, opened 2022-05-24
feat: add v37 support for developers

Jack-Works:
A new option (only visible in development mode or beta/nightly channel) to use v-37 payload.

image

Notable changes:

- No image payload support yet.
- Now you can use **encrypt to self** with no local key (which is not possible in v-38).
- Payload is shorter.

V38 payload:

image

V37 payload: 

image


Jack-Works:
Merge conflict fixed.

PR #6592, opened 2022-06-15
fix(tips): share receiver status via storage

UncleBill:


UncleBill:
Close, since it doesn't fix the root cause.

PR #6676, opened 2022-07-04
fix: show nft avatar for solana

lelenei:
Closes #  mf-1106


Jack-Works (approved):

guanbinrui (approved):

Jack-Works:
I believe there is something wrong with the type definition, otherwise these kinds of errors can be caught by the type checker.

PR #6717, opened 2022-07-07
fix: temporarily ignore ts error

UncleBill:
Need @Randolph314  to refactor some code.


UncleBill:
closes as completed in https://github.com/DimensionDev/Maskbook/pull/6716

PR #6743, opened 2022-07-11
fix(trending): mf-1437 layout of buy button

UncleBill:
https://mask.atlassian.net/browse/MF-1437


UncleBill:
Has been verified by QA.

PR #6751, opened 2022-07-11
fix: replace web3 to connection

nuanyang233:
- fix: replace web3 to connection in trader plugin

## Description

Closes # (NO_ISSUE)

## Type of change

- [ ] Documentation

## Checklist


packages/mask/src/plugins/Trader/trader/0x/useTradeCallback.ts
```diff
@@ -27,28 +27,23 @@ export function useTradeCallback(tradeComputed: TradeComputed<SwapQuoteResponse>
...
             ...config,
-            gas: await web3.eth
-                .estimateGas({
+            gas:
+                (await connection.estimateTransaction?.({
                     from: account,
                     ...pick(tradeComputed.trade_, ['to', 'data', 'value']),
-                })
-                .catch(() => 0),
+                })) ?? ZERO.toString(),
```
UncleBill, line 40: If we send with `connection.sendTransaction`, then we can omit `gas` here.

@guanbinrui  can confirm this.

But, here is another question: do we need to add `data` field at https://github.com/DimensionDev/Maskbook/blob/1f78d98466926d75bdceb45339eabfa6ea7e797f/packages/web3-shared/evm/utils/contract.ts#L57-L62

```diff
diff --git a/packages/web3-shared/evm/utils/contract.ts b/packages/web3-shared/evm/utils/contract.ts
index db5966518..83108f2c2 100644
--- a/packages/web3-shared/evm/utils/contract.ts
+++ b/packages/web3-shared/evm/utils/contract.ts
@@ -57,6 +57,7 @@ export async function encodeContractTransaction(
     if (!tx.gas) {
         tx.gas = await transaction.estimateGas({
             from: tx.from as string | undefined,
+            data: tx.data,
             value: tx.value,
         })
     } 
```

packages/mask/src/plugins/Trader/trader/bancor/useTradeCallback.ts
```diff
@@ -36,20 +35,15 @@ export function useTradeCallback(tradeComputed: TradeComputed<SwapBancorRequest>
         const config = pick(tradeTransaction.transaction, ['to', 'data', 'value', 'from'])
         const config_ = {
             ...config,
-            gas: await web3.eth.estimateGas(config).catch((error) => {
-                throw error
-            }),
+            gas: (await connection.estimateTransaction?.(config)) ?? ZERO.toString(),
```
UncleBill, line 38: Ditto, and don't forget to check other places in this patch.

packages/mask/src/plugins/Trader/trader/0x/useTradeCallback.ts
nuanyang233: No, `connection.sendTransaction` does not automatically add `gas`.
 
It uses `encodeTransaction` instead of `encodeContractTransaction`.

 But it needs to add `data` and `to` field to estimate gas. @guanbinrui will add it.

PR #6780, opened 2022-07-12
fix: avatar chain tab

lelenei:
image


lelenei:
ready fixed

lelenei:
ready fixed

PR #6807, opened 2022-07-15
refactor: icons codegen

### LINKED ISSUE #3563: [Demand] Codegen for @masknet/icons package

Jack-Works:
close MF-286
close #5912
close #3563
https://mask.atlassian.net/browse/MF-284


packages/icons/convert.js
```diff
@@ -0,0 +1,99 @@
+import { readdir, readFile, unlink, writeFile } from 'node:fs/promises'
```
UncleBill, line 1: Don't forget to remove this migration script before merging.

.gitignore
```diff
@@ -1,7 +1,6 @@
 # generated files
-i18n_generated.js
-i18n_generated.d.ts
-i18n_generated.d.ts.map
+i18n_generated*
+icon-generated*
```
UncleBill, line 3: I don't think it's good to ignore these generated files. Ignoring them makes switch among branches painful.

They are generated by tools indeed, but they are also source code. I would prefer our code can run without building  scripts.

Beyond that, we can also track the **real** changes.

UncleBill:
Following files have two layers of `<svg />`, please fix them.
image

.gitignore
Jack-Works: if we do this, we should also track i18n generated files. We can discuss that and (if we decided to) make a new PR to track both icons generated and i18n generated.

I'll not do that in this PR.

PR #6895, opened 2022-07-27
refactor: migrate icons

UncleBill:
`<XXXIcon />` to  `<Icons.XXX />`


packages/scripts/src/codegen/icon-codegen.ts
```diff
@@ -165,9 +164,7 @@ async function generateIcons() {
 
         // export const TIcon: ...
         if (!Ident.endsWith('Icon')) {
```
Jack-Works: you should remove this whole if statement

Jack-Works (approved):

PR #6915, opened 2022-07-29
fix: update host of NFTScan's API

UncleBill:
https://mask.atlassian.net/browse/MF-1614

### DIFF
--- a/packages/mask/src/utils/fetcher.ts
+++ b/packages/mask/src/utils/fetcher.ts
@@ -6,13 +6,13 @@ export function contentFetch(input: RequestInfo, init?: RequestInit) {
 
     if (isSameOrigin(info.url)) {
         if (process.env.engine === 'firefox' && process.env.manifest === '2' && typeof content === 'object') {
-            return content.fetch(info)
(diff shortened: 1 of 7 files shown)


packages/mask/src/utils/fetcher.ts
```diff
@@ -6,13 +6,13 @@ export function contentFetch(input: RequestInfo, init?: RequestInit) {
...
-            return content.fetch(info)
+            return content.fetch(info, init)
         } else {
-            return original_fetch(info)
+            return original_fetch(info, init)
         }
     }
 
-    return Services.Helper.r2d2Fetch(info)
+    return Services.Helper.r2d2Fetch(info, init)
```
UncleBill, line 15: `Request.headers` can't be serialized.

guanbinrui (approved):

packages/mask/src/utils/fetcher.ts
Jack-Works: oh, maybe we can add one. it is not hard

PR #6918, opened 2022-07-29
fix: backdrop-filter on FF 103

UncleBill:


guanbinrui (approved):

Jack-Works:
@guanbinrui as we discussed above, those fixes should be merged into develop then picked into release branches.

PR #6932, opened 2022-07-31
chore: optimize svg with svgo

### LINKED ISSUE #4222: [Bug] SVG id conflicts

UncleBill:
closes #4222


packages/icons/brands/Blocto.svg
```diff
@@ -1,43 +1,21 @@
...
-                values="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 127 0"
-                result="hardAlpha"
-            />
-            <feOffset dy="6" />
-            <feGaussianBlur stdDeviation="6" />
-            <feComposite in2="hardAlpha" operator="out" />
-            <feColorMatrix type="matrix" values="0 0 0 0 0.203922 0 0 0 0 0.521569 0 0 0 0 0.768627 0 0 0 0.2 0" />
-            <feBlend mode="normal" in2="effect1_backgroundBlur_270_89338" result="effect2_dropShadow_270_89338" />
-            <feBlend mode="normal" in="SourceGraphic" in2="effect2_dropShadow_270_89338" result="shape" />
+        <filter id="Blocto_svg__a" width="62" height="64" x="-9.5" y="-12" color-interpolation-filters="sRGB" filterUnits="userSpaceOnUse">
```
UncleBill, line 8: `svgo` prefix the id, that how we resolve #4222

package.json
```diff
@@ -23,6 +23,7 @@
     "lint:ci-report": "pnpm run lint:ci --format=junit --output-file=reports/junit/eslint-results.xml --max-warnings=0",
     "postinstall": "gulp fix-lockfile",
     "prepare": "husky install",
+    "svgo": "svgo packages/icons/*/*.svg",
```
Jack-Works, line 26: please install it in packages/scripts or packages/icons.

packages/mask/.webpack/config.ts
```diff
@@ -136,6 +137,20 @@ export function createConfiguration(rawFlags: BuildFlags): Configuration {
                         },
                     },
                 },
+                {
+                    test: /\.svg$/,
+                    use: [
+                        {
+                            loader: 'svgo-loader',
```
Jack-Works: please don't add loaders to convert non-js files into a js file.

and it seems like you doesn't use it, so you don't need to add this.

package.json
UncleBill: No, need it to run in git hook. Installing in root directory would be better.

packages/mask/.webpack/config.ts
UncleBill: We only refer svg with `new URL()`, in that way webpack won't convert it into JS but file.

It's to minify svg for release.

packages/mask/.webpack/config.ts
```diff
@@ -136,6 +137,22 @@ export function createConfiguration(rawFlags: BuildFlags): Configuration {
                         },
                     },
                 },
+                mode === 'production'
+                    ? {
+                          test: /\.svg$/,
+                          use: [
+                              {
+                                  loader: 'svgo-loader',
+                                  // overrides
```
Jack-Works:
```suggestion
                                  loader: 'svgo-loader',
                                  dependency: (data) => {
                                      if (data === "") return false
                                      if (data !== 'url') throw new TypeError('The only import mode valid for a non-JS file is via new URL(). Current import mode: ' + data)
                                      return true
                                  },
                                  type: 'asset/resource'
```

package.json
Jack-Works: I'm wondering, if you're running it in the source code, why you need to optimize it again in webpack?

UncleBill: Formatting in git hook, minifying in building.

UncleBill: The loader is not enabled in development mode.

Jack-Works: that's interesting

Jack-Works (approved):

PR #6957, opened 2022-08-04
feat: add  background page

lelenei:
Closes mf-702


packages/mask/src/plugins/Avatar/Application/NFTListPage.tsx
```diff
@@ -141,29 +150,6 @@ export function NFTImageCollectibleAvatar({
...
-        <img
-            className={classes.image}
-            src={
-                theme.palette.mode === 'dark'
-                    ? assetPlayerFallbackImageDark.toString()
-                    : assetPlayerFallbackImageLight.toString()
-            }
-        />
-    )
+    return
```
guanbinrui: Why not remove it?

packages/mask/src/utils/components/WalletStatusBar/PluginVerifiedWalletStatusBar.tsx
```diff
@@ -98,10 +98,12 @@ export const PluginVerifiedWalletStatusBar = memo<PluginVerifiedWalletStatusBarP
                 providerIcon: account ? providerDescriptor?.icon : undefined,
                 iconFilterColor: account ? providerDescriptor?.iconFilterColor : '',
                 formattedAddress: Others?.formatAddress(account || (defaultVerifiedWallet?.identity ?? ''), 4),
-                addressLink: Others?.explorerResolver.addressLink?.(
-                    account ? chainId : defaultChainId,
-                    account || (defaultVerifiedWallet?.identity ?? ''),
-                ),
+                addressLink:
```
nuanyang233, line 101: Please revert this change. `addressLink` should be a URL and should not use account.

nuanyang233 (changes requested):

packages/mask/src/utils/components/WalletStatusBar/PluginVerifiedWalletStatusBar.tsx
lelenei: you tested?

nuanyang233:
```
 <Link
          href={addressLink}
          target="_blank"
          title="View on Explorer"
          rel="noopener noreferrer"
          onClick={(event) => {
                 event.stopPropagation()
           }}
          className={classes.linkIcon}>
         <Icons.LinkOut size={14} className={classes.linkIcon} />
</Link>
```

So you want to set an address to `href`? Do you think that makes sense?

nuanyang233: And test result: 
image

image

guanbinrui (approved):

PR #7164, opened 2022-08-20
ci: don't upload zip artifacts with upload-artifact

UncleBill:
https://github.com/actions/upload-artifact#zipped-artifact-downloads

image


UncleBill:
GitHub doesn't show the compressed size.
image

Jack-Works:
I tried the same thing at #7049, but I think we cannot do that. Because it counts each file upload as an API request, and if we have too many API requests (about 1000~ times every run), we will hit the rate limit very soon.

PR #7257, opened 2022-08-31
fix(NFTScan): mf-1895 correct collection data normalization

UncleBill:
https://mask.atlassian.net/browse/MF-1895

image

### DIFF
--- a/packages/web3-providers/src/NFTScan/helpers/EVM.ts
+++ b/packages/web3-providers/src/NFTScan/helpers/EVM.ts
@@ -80,7 +80,11 @@ export function createPermalink(chainId: ChainId, address: string, tokenId: stri
     })
 }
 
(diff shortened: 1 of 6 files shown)


packages/web3-providers/src/NFTScan/helpers/EVM.ts
```diff
@@ -134,16 +139,18 @@ export function createNonFungibleAsset(chainId: ChainId, asset: EVM.Asset): NonF
...
         collection: {
             chainId,
-            name,
+            name: contractName,
             slug: name,
             description,
             address: asset.contract_address,
-            iconURL: urlcat(NFTSCAN_LOGO_BASE + '/:id', { id: asset.contract_address + '.png' }),
+            iconURL: collectionContext
+                ? collectionContext.logo_url
```
UncleBill: If `collectionContext.logo_url` is null, we will render a fallback logo instead directly.

guanbinrui (approved):

PR #7269, opened 2022-09-01
fix: fails to get SocialAddressList for Avatar

UncleBill:
https://mask.atlassian.net/browse/MF-1920


guanbinrui (approved):

UncleBill:
https://mask.atlassian.net/browse/MF-1920

PR #7299, opened 2022-09-06
fix(ProfileCard): handle edge case and improve ux

UncleBill:
https://mask.atlassian.net/browse/MF-1943


guanbinrui (approved):

packages/mask/src/plugins/ProfileCard/SNSAdaptor/ProfileAvatarBadge.tsx
```diff
@@ -14,30 +14,32 @@ const useStyles = makeStyles()((theme) => ({
 interface Props extends IconButtonProps {
     userId: string
 }
+let closeTimer: NodeJS.Timeout
```
Jack-Works, line 17: please explain why you need to store info outside of a react component.

PR #7640, opened 2022-10-11
feat: mf-2232 allow to Tip multiple times in a single TipsDialog

UncleBill:
https://mask.atlassian.net/browse/MF-2232


guanbinrui (approved):

packages/mask/src/social-network/define.ts
```diff
@@ -7,7 +7,17 @@ export const definedSocialNetworkUIs: ReadonlyMap<string, SocialNetworkUI.Deferr
...
+    return import('./ui.js')
+        .then((x) => x.activateSocialNetworkUIInner(ui_deferred))
+        .then(
+            () => true,
+            (error) => {
+                if (process.env.NODE_ENV === 'development') {
+                    console.info('Failed to initialize social network UI', error)
+                }
+                return false
+            },
```
Jack-Works, line 19: is this happening? in which case?

nuanyang233: I think this is from my bug, which has been fixed in develop

PR #7700, opened 2022-10-17
fix(Solana): mf-2444 address validation is too strict

UncleBill:
https://mask.atlassian.net/browse/MF-2444

### DIFF
--- a/packages/web3-shared/solana/src/utils/address.ts
+++ b/packages/web3-shared/solana/src/utils/address.ts
@@ -17,7 +17,7 @@ export function decodeAddress(initData: string | Buffer) {
 }
 
 export function formatAddress(address: string, size = 0) {
-    if (!isValidAddress(address)) return address
+    if (!isValidAddress(address, false)) return address
     if (size === 0 || size >= 22) return address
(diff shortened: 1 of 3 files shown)


packages/web3-shared/solana/src/utils/address.ts
```diff
@@ -43,7 +43,7 @@ export function isValidAddress(address?: string) {
     const length = address?.length
     if (!length) return false
     try {
-        return length >= 32 && length <= 44 && Web3.PublicKey.isOnCurve(bs58.decode(address))
```
guanbinrui, line 46: Strangely, even the Phantom extension cannot accept your given address.

image

I did some research about it. Found  more such kind addresses:

+ `4ig5eTKvBefXB9Pqwq2rCJTGFRNATEa7geNzeuVQGmAE`
+ `Eej78Z5E4w5sSreb8w8esQqTVwBuyT8La2ypa8jiY3Tw`

A Solana explorer page about each of those addresses could be found. However, the `isOnCurve` method returns `false.`

guanbinrui (approved):

PR #7871, opened 2022-11-03
fix: mf-2618 result of useId can't be used as svg element id

UncleBill:
https://mask.atlassian.net/browse/MF-2618

a488102f4e358c402f3


Jack-Works:
that surprise me, since useId is used for HTML ID

PR #7941, opened 2022-11-10
chore: upgrade to ts 5.0

Jack-Works:
The compiler is now 10-25% faster. tsc is now 30% faster to start. npm package is now 43% smaller.

see https://github.com/microsoft/TypeScript/pull/51387


Jack-Works:
looks like there is a small breaking change that fails `@storybook/react-docgen-typescript-plugin`. I have reported to typescript, let's wait.

https://github.com/microsoft/TypeScript/issues/51473

PR #7969, opened 2022-11-11
fix: ssr crash

Jack-Works:
image

@UncleBill the Markdown component you added in the theme package crashes SSR. I added a fix to lazy load those libraries, please verify if it work normally thanks!

### DIFF
--- a/packages/theme/src/Components/Markdown/index.tsx
+++ b/packages/theme/src/Components/Markdown/index.tsx
@@ -1,13 +1,15 @@
 import { markdownTransformIpfsURL } from '@masknet/shared-base'
-import { memo } from 'react'
-import ReactMarkdown, { Options } from 'react-markdown'
-import remarkGfm from 'remark-gfm'
-import rehypeRaw from 'rehype-raw'
+import { memo, use, lazy } from 'react'
+import type {} from 'react/next'
+import type { Options } from 'react-markdown'
+export type { Options } from 'react-markdown'
 
-interface Props extends Options {}
-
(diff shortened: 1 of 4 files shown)


packages/theme/src/Components/Markdown/index.tsx
```diff
@@ -1,13 +1,15 @@
...
+import type { Options } from 'react-markdown'
+export type { Options } from 'react-markdown'
 
+// Note: this 3 dependencies crashes in background script. we need to lazy load it until use.
+const ReactMarkdown = lazy(() => import('react-markdown'))
 // TODO Succeed another Markdown in '@masknet/shared'
-export const Markdown = memo<Props>(({ children, ...props }) => {
+export const Markdown = memo<Options>(function Markdown({ children, ...props }) {
+    const remarkGfm = use(import('remark-gfm').then((x) => x.default))
+    const rehypeRaw = use(import('rehype-raw').then((x) => x.default))
```
UncleBill, line 12: Can we move them out of the component? Lazy loading inside causes UI to flicker.

You can checkout RSS3 feed cards like this one, and click to open the detail dialog.

image

Jack-Works: you can write `import('remark-gfm')` in RSS3 plugin's SNS Adaptor's `init()` method to pre-load it.

UncleBill: I don't think preloading would resolve it. The flicker happens every time the dialog opens.

PR #8220, opened 2022-12-08
fix: mf-2951 calculate file entry digest including upload options

UncleBill:
https://mask.atlassian.net/browse/MF-2951
https://mask.atlassian.net/browse/MF-2952

### DIFF
--- a/packages/plugins/FileService/src/types.ts
+++ b/packages/plugins/FileService/src/types.ts
@@ -33,11 +33,19 @@ export interface ProviderAgent {
 export interface FileInfo {
     type: 'file'
     provider: Provider
+    /**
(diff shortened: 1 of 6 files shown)


guanbinrui (approved):

packages/plugins/FileService/src/types.ts
```diff
@@ -33,11 +33,19 @@ export interface ProviderAgent {
...
     id: string
 
     name: string
     size: number
-    createdAt: Date
+    /**
+     * Before v3, it's Date
+     * Since v3, it's number
+     */
+    createdAt: number
```
Jack-Works, line 48: why not use Date?

PR #8243, opened 2022-12-12
chore: make sure pnpm lock file is always up to date

UncleBill:


UncleBill:
Cancel

Jack-Works:
isn't this being checked by `pnpm install` on CI? 🤔

UncleBill:
> isn't this being checked by `pnpm install` on CI? 🤔

I think `pnpm install` on CI will respect to lock file by default. `pnpm` it self can detect CI environment.

PR #8264, opened 2022-12-14
fix: render text has duplicate keys

UncleBill:
fix annoying warnings

https://mask.atlassian.net/browse/MF-2989


guanbinrui (approved):
Thanks for fixing those warnings.

PR #8513, opened 2023-01-16
fix(Tips): mf-3171 switch network by selected recipient

UncleBill:
https://mask.atlassian.net/browse/MF-3171

### DIFF
--- a/packages/mask/src/plugins/Tips/contexts/Tip/TipTaskProvider.tsx
+++ b/packages/mask/src/plugins/Tips/contexts/Tip/TipTaskProvider.tsx
@@ -3,9 +3,9 @@ import { useSubscription } from 'use-subscription'
 import {
     useFungibleToken,
     useNonFungibleTokenContract,
-    useNetworkContext,
     useChainContext,
     getDefaultChainId,
+    useNetworkContext,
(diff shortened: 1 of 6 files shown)


packages/mask/src/plugins/Tips/contexts/Tip/TipTaskProvider.tsx
```diff
@@ -59,11 +59,18 @@ export const TipTaskProvider: FC<React.PropsWithChildren<Props>> = memo(({ child
...
-    const token = userSelectedToken ?? nativeTokenDetailed
-    useEffect(() => {
-        setToken(nativeTokenDetailed)
-    }, [nativeTokenDetailed])
+    const [tokenMap, setTokenMap] = useState<Record<string, TipContextOptions['token']>>({})
+    const setToken: TipContextOptions['setToken'] = useCallback(
+        (val) => {
+            const key = `${pluginID}:${chainId}`
+            setTokenMap((map) => {
+                const newToken = typeof val === 'function' ? val(map[pluginID]) : val
```
UncleBill:
```suggestion
                const newToken = typeof val === 'function' ? val(map[key]) : val
```

guanbinrui (approved):

PR #8568, opened 2023-01-30
fix: fixup mf-3278 adjust icons

UncleBill:
https://mask.atlassian.net/browse/MF-3278


guanbinrui (approved):

packages/dashboard/src/pages/routes.tsx
```diff
@@ -39,11 +40,17 @@ export function Pages() {
...
     }, [restoreCallback])
-
     return (
         <Suspense fallback={null}>
             <Routes>
-                <Route path={DashboardRoutes.Welcome} element={<Welcome />} />
+                <Route
+                    path={DashboardRoutes.Welcome}
+                    element={
+                        <NoPersonaGuardRoute redirectTo={DashboardRoutes.Personas}>
```
Jack-Works, line 49: is this change breaks https://github.com/DimensionDev/Maskbook/pull/8328?

UncleBill: No, it shouldn't be removed in #8328. Removing it is for bypassing.

PR #8656, opened 2023-02-09
fix: mf-3343 storage agreed state

UncleBill:
https://mask.atlassian.net/browse/MF-3343


guanbinrui (approved):

packages/dashboard/src/hooks/useTermsAgreed.ts
```diff
@@ -0,0 +1,16 @@
+import { useCallback, useState } from 'react'
+import { createContainer } from 'unstated-next'
+
+const KEY = 'dashboard/terms-agreed'
```
Jack-Works, line 4: terms should have a version. e.g. `term-2023-01`, otherwise we don't know if the user accepts the latest term or not

PR #8670, opened 2023-02-10
fixup! fix: mf-3342 handle Translate values contain tag-like string (…

UncleBill:
https://mask.atlassian.net/browse/MF-3342


guanbinrui (approved):

packages/plugins/RSS3/src/SNSAdaptor/components/FeedCard/CollectibleCard.tsx
```diff
@@ -132,6 +132,7 @@ export const CollectibleCard: FC<CollectibleCardProps> = ({ feed, ...rest }) =>
                     metadata,
                     summary: (
                         <Translate.collectible_mint
+                            shouldUnescape
```
UncleBill, line 135: @Jack-Works Could you update the i18n generator, and add `shouldUnescap: true` to the wrapped `<Trans />`?

```ts
function createComponent(i18nKey: string) {
    return (props) => createElement(Trans, { i18nKey, ns: "com.maskbook.fileservice", shouldUnescape: true, ...props });
}
```

Jack-Works: OK, I'll do that

Jack-Works: @UncleBill please try 0.5.0, a new option in the generator `shouldUnescape` is added

PR #8784, opened 2023-02-23
docs(i18n): sync i18n files from Crowdin

Jack-Works:


packages/mask/shared-ui/locales/ja-JP.json
```diff
@@ -149,7 +149,7 @@
     "plugin_ito_dialog_swap_exceed_wallet_limit": "一つのウォレットの限度を超えています",
     "plugin_ito_swap_ration_label": "スワップ比率",
     "plugin_ito_swap_only_once_remind": "各ウォレットは一度しか参加できません。",
-    "plugin_ito_swap_title": "{{swap}} 分の {{- amount}} {{token}}",
+    "plugin_ito_swap_title": "{{swap}} 分の {{amount}} {{token}}",
```
Jack-Works, line 152: this should be added in crowdin, otherwise, it might be overwritten when sync with crowdin (maybe cc @UncleBill)

UncleBill: Yes. I think this patch shouldn't be merged until we get them fixed in Crowdin.

PR #8792, opened 2023-02-25
chore: infer class names from useStyles() when override classes is not provided

UncleBill:
close MF-0000

## Before

image

## After

image

### DIFF
--- a/packages/theme/src/UIHelper/makeStyles.ts
+++ b/packages/theme/src/UIHelper/makeStyles.ts
@@ -22,7 +22,7 @@ export const { makeStyles } = createMakeStyles({ useTheme }) as any as {
         },
     ) => {
         classes: StyleOverrides extends { classes?: { [key in infer ExtraKeys]?: string | undefined } }
-            ? Record<ExtraKeys | RuleName, string>
+            ? Record<string extends ExtraKeys ? RuleName : ExtraKeys | RuleName, string>
             : Record<RuleName, string>
         theme: Theme
         css: Css
--- a/packages/mask/src/components/InjectedComponents/SearchResultInspector.tsx
+++ b/packages/mask/src/components/InjectedComponents/SearchResultInspector.tsx
@@ -90,12 +90,12 @@ export function SearchResultInspector(props: SearchResultInspectorProps) {
     if (!contentComponent) return null
 
     return (
-        <div className={classes.root}>
+        <div>
             <ScopedDomainsContainer.Provider>
                 <div className={classes.contentWrapper}>
-                    <div className={classes.content}>{contentComponent}</div>
+                    <div>{contentComponent}</div>
                     {tabs.length ? (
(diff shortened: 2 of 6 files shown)


packages/theme/src/UIHelper/makeStyles.ts
```diff
@@ -22,7 +22,7 @@ export const { makeStyles } = createMakeStyles({ useTheme }) as any as {
         },
     ) => {
         classes: StyleOverrides extends { classes?: { [key in infer ExtraKeys]?: string | undefined } }
-            ? Record<ExtraKeys | RuleName, string>
+            ? Record<string extends ExtraKeys ? RuleName : ExtraKeys | RuleName, string>
```
Jack-Works, line 25: can you explain this change? It looks like `string extends ExtraKeys` always be true, therefore it will always go to the truthy branch.

UncleBill: > It looks like string extends ExtraKeys always be true

If `StyleOverrides` is `{}`, then `ExtraKeys` would not be an `union` but a `string`, and `string extends ExtraKeys` would be `true`

image


https://www.typescriptlang.org/play?#code/C4TwDgpgBAygjFAvFA3lAxgGwIYGdcS4BcqUA1iQOQBulUAvgwFCiSwBMSq9TL40AaQTJ4UCAA9gEAHYATXKSx4CuAPwk0AbTIQQUAJbSD0gGYQATlACik89gG7cAXXVRcwc4YDmUAD5QAVzkIE0MIWQZI1Wtbe0coEmkIagsmKCgAegyoAD1VPjYBThFOCSk5BTQlfEJXLR09Q2MzSxsPOJBnV3dPaR9-INkQsIjGRmi2uwdOhKgklPM0zOy8pazYUEwIAHkFzyGFMpl5RRwatQ0obV1jNw9vFxIe7z9A4NCk0YYAGjcAexi7WmCn0CmefQK0AASlwiq9KNVcHBKGsVvleEwsljsTjcXj8QT8ZCNuYACr8LjgrzEgCq0n05LYyEoAAkIJhMH8UcSYKIRB5GdAjhU7r0fNFKCBCHQqNIuajcvlWNBecVRS9hScYAKKRKpbgZVBKHKUel1qteMqoDS+db6YKxJJjmD7n0oHrpbNjfKzWjaTA1VTHeUTnSGbqjfrDd7TctFUA

UncleBill: > It looks like string extends ExtraKeys always be true

`string extends 'union' | 'type'`  would be `false`

UncleBill: @Jack-Works Any other questions?

Jack-Works (approved):

guanbinrui (approved):

PR #8804, opened 2023-02-27
chore: upgrade infra deps

Jack-Works:
MF-0000


guanbinrui (dismissed):

Jack-Works:
this PR does not affect runtime code so I'll merge this soon.

I rewrite the eslint rule set, most of the new rules are disabled and I'm intended to enable them case-by-case in the following PR>

PR #9113, opened 2023-03-20
fix(RSS3): mf-3313 some small issues

UncleBill:
https://mask.atlassian.net/browse/MF-3313


packages/web3-providers/src/types/RSS3.ts
```diff
@@ -189,13 +189,11 @@ export namespace RSS3BaseAPI {
...
-            {
-                /** URL */
-                address: string
-                mime_type: 'image/png' | string
-            },
-        ]
+        media: Array<{
+            /** URL */
+            address: string
+            mime_type: 'image/png' | string
```
guanbinrui, line 195: Make the editor happy.

```suggestion
            mime_type: 'image/png' | Omit<string, 'image/png'>
```

UncleBill: Got such error:
image

Jack-Works: `'image/png' | string` is equal to `string`

guanbinrui: > Got such error:
> 
> image
> 
> 

Omit will help auto completion. Maybe cast it to string?

UncleBill: > `'image/png' | string` is equal to `string`

I know, just want to document it in code.

UncleBill: Would like to list as many types as possible.

PR #9241, opened 2023-04-04
fix: adjust tips and lens UI

UncleBill:
https://mask.atlassian.net/browse/MF-3837
https://mask.atlassian.net/browse/MF-3875
https://mask.atlassian.net/browse/MF-3877
https://mask.atlassian.net/browse/MF-3898

### DIFF
--- a/packages/shared/src/UI/components/Markdown/index.tsx
+++ b/packages/shared/src/UI/components/Markdown/index.tsx
@@ -1,11 +1,14 @@
 /// <reference types="@masknet/global-types/dom" />
 /// <reference types="react/next" />
 import { markdownTransformIpfsURL } from '@masknet/shared-base'
-import { memo, use } from 'react'
+import { memo } from 'react'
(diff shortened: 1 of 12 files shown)


guanbinrui (approved):

packages/shared/src/UI/components/Markdown/index.tsx
```diff
@@ -39,9 +42,6 @@ interface MarkdownProps extends Options {
 
 export const Markdown = memo<MarkdownProps>(({ children, className, defaultStyle = true, ...props }) => {
     const { classes, cx } = useStyles()
-    const { default: rehypeRaw } = use(import('rehype-raw'))
```
Jack-Works, line 42: why？

UncleBill: Will cause UI flickering, see https://mask.atlassian.net/browse/MF-3877

PR #9430, opened 2023-04-21
fix: mf-4044 solana link

UncleBill:
https://mask.atlassian.net/browse/MF-4044


packages/plugins/Collectible/src/SNSAdaptor/List/Collection.tsx
```diff
@@ -136,7 +136,7 @@ export const Collection: FC<CollectionProps> = memo(
                 className={classes.collectibleCard}
                 asset={asset}
                 pluginID={pluginID}
-                key={`${collection.id}.${asset.tokenId}`}
+                key={`${collection.id}.${asset.address}.${asset.tokenId}`}
```
UncleBill, line 139: Solana NFTs could be duplicate.

guanbinrui (approved):

PR #9455, opened 2023-04-26
chore: explicitly sort icon files

UncleBill:
MF-0000
To avoid uncertain results.


guanbinrui (approved):

packages/scripts/src/codegen/icon-codegen.ts
```diff
@@ -101,36 +101,38 @@ async function generateIcons() {
...
-        currentLine += `${identifier}_url: string`
-        asURL.dts.push(currentLine)
-
-        const isDynamicColor = sourceMap.has(path)
-        const source = isDynamicColor ? sourceMap.get(path)! : null
-        variants[base].push({
-            args: [currentVariant, url, source, isDynamicColor],
-            assetPath: importPath,
+    filePaths
+        .sort((a, z) => a.localeCompare(z))
```
Jack-Works, line 105: locale compare is by default using the host's settings, so you should explicitly use `localeCompare(z, 'en-us')`. I will fix this in my PR

PR #9457, opened 2023-04-26
chore: optimize icons codegen output

Jack-Works:
MF-0000

Optimize for speed and DX.

Before:

image

Now: 
image

### DIFF
--- a/packages/icons/utils/internal.js
+++ b/packages/icons/utils/internal.js
@@ -8,7 +8,7 @@ import { MaskIconPaletteContext } from './MaskIconPaletteContext.js'
 
 /**
  * @param {string} name
- * @param {Array<RawIcon>} variants
+ * @param {Array<import('./internal.js').__RawIcon__>} variants
  * @param {[number, number]} intrinsicSize
  * @returns {React.ComponentType<import('./internal').GeneratedIconProps>}
  */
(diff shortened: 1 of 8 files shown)


UncleBill:
Why?

We don't need to regenerate the code if we track them in the codebase. Or it'll be painful when you switch among beaches or syncing code that has icons added or removed.

packages/icons/utils/internal.js
```diff
@@ -1,14 +1,11 @@
...
 import { MaskIconPaletteContext } from './MaskIconPaletteContext.js'
 
-/**
- * @typedef {[currentVariant: null | string[], url: string, jsx: object | null, supportColor?: boolean]} RawIcon
- */
-
 /**
  * @param {string} name
- * @param {Array<RawIcon>} variants
+ * @param {Array<import('./internal.js').__RawIcon__>} variants
```
UncleBill, line 11: I think the previous structure would be cleaner.

Jack-Works:
> Why?
> 
> We don't need to regenerate the code if we track them in the codebase. Or it'll be painful when you switch among beaches or syncing code that has icons added or removed.

Currently, there are 3 problems:

1. Developers might forget to run it and commit the code.
2. It might be not reproducible on different machines (e.g. file name order).
3. `.d.ts` files should be also checked in otherwise there will be type errors.

If you think we still should check in those files, I can resolve it by

- Add a new CI check and fails if the dev forgets to run it to the latest.
- Make sure it is reproducible.
- Also check in `.d.ts` files.

packages/icons/utils/internal.js
Jack-Works: This new structure is intended to lazy load and make it smaller. The tuple structure does not allow me to omit some fields.

guanbinrui (dismissed):

PR #9471, opened 2023-04-27
fix: mf-4069 wip ens name wrapper

UncleBill:
https://mask.atlassian.net/browse/MF-4069


UncleBill:
Fixed by SimpleHash the API provider.

PR #9543, opened 2023-05-08
fix: mf-4147 remove outdated translation

UncleBill:
https://mask.atlassian.net/browse/MF-4147


guanbinrui (approved):

UncleBill:
@guanbinrui Could you sync this change to develop branch which hosts the remote translation that would overrides the builtin translation at runtime.

PR #9548, opened 2023-05-09
fix: mf-4149 workaround for bad Gticoin url

UncleBill:
https://mask.atlassian.net/browse/MF-4149


UncleBill:
We don't need this workaround if the url can be fixed before our release. So just keep drafting for now.

UncleBill:
Has been fixed by RSS3 team.

PR #9581, opened 2023-05-11
refactor: mf-4157 query data with react-query

UncleBill:
https://mask.atlassian.net/browse/MF-4157


guanbinrui (approved):
LGTM

package.json
```diff
@@ -41,6 +41,7 @@
     "@mui/lab": "5.0.0-alpha.102",
     "@mui/material": "5.10.8",
     "@mui/system": "5.10.8",
+    "@tanstack/react-query": "^4.29.5",
```
Jack-Works, line 44: should not install them in the root?

PR #9623, opened 2023-05-17
perf: optimize startup time

Jack-Works:
- perf: defer wallet.ts
- perf: defer anchorme
- chore: defer blocto/fcl
- chore: defer api creation
- chore: defer @metamask/eth-sig-util
- chore: optimize removeEmptyString
- chore: optimize theme pkg
- chore: optimize pkgs

This is the split PR.

MF-3826


Jack-Works:
this pr is gradually split into many PRs.

PR #9679, opened 2023-05-26
feat: improve devtools

Jack-Works:
close mf-4204

### Persistent editor URI

See `.env/dev-preference.example`.

You can use this `.env/dev-preference` to keep your preferred editor URI in a new install.

```env
### React devtools default editor

devtoolsEditorURI=vscode://file/{path}:{line}
# (default) VSCode

# devtoolsEditorURI=...
# Any URI that can open the editor you used.
```

### Reload and Profile

Now enabled, but might not work properly due to a likely upstream bug.

### Enabled on Firefox

Known issue:

- Cannot use Reload and Profile (upstream issue)
- Not work when the inspector panel is not docked (on it's own window)
- Open in Editor does not work

### Others

- Now preferences are correctly stored.
- Fix Manifest V3 crashes
- Improve UI when the page refreshes


packages/mask/devtools/panels/utils.ts
```diff
@@ -1,20 +1,24 @@
...
+            else throw new Error((exception as any).code)
+        }
+        return result as T
     }
-    return result
 }
 
 export function createPanel(name: string) {
-    return browser.devtools.panels.create(name + ' (\u{1F3AD})', '128x128.png', 'empty.html')
+    return browser.devtools.panels.create(name + ' (Mask)', '128x128.png', 'empty.html')
```
UncleBill:
```suggestion
    return browser.devtools.panels.create(name, '128x128.png', 'empty.html')
```
We can distinguish ours and others by emoji.

image

packages/mask/.webpack/config.ts
```diff
@@ -165,12 +165,13 @@ export function createConfiguration(_inputFlags: BuildFlags): Configuration {
             (() => {
                 // In development mode, it will be shared across different target to speedup.
                 // This is a valuable trade-off.
-                const runtimeValues = {
+                const runtimeValues: any = {
```
guanbinrui: Avoid using any.

packages/mask/devtools/panels/index.tsx
```diff
@@ -3,12 +3,17 @@ import { attachListener } from './utils.js'
...
     const controller = new AbortController()
 
-    // this code runs in a invisible page, therefore it's requestAnimationFrame is not being called.
-    // some UI libraries (like react-window) require that to work.
-    globalThis.requestAnimationFrame = function (callback) {
-        return requestIdleCallback((x) => callback(x.timeRemaining()))
+    // See https://github.com/facebook/react/pull/26193
+    const FRAME_TIME = 16
+    let lastTime = 0
+    globalThis.requestAnimationFrame = function (callback: any) {
```
guanbinrui: Avoid using any.

guanbinrui (dismissed):

guanbinrui (approved):

PR #9821, opened 2023-06-08
fix: mf-4301 defer checking click away, setting click inside tree first

UncleBill:
https://mask.atlassian.net/browse/MF-4301


Jack-Works:
If a patch has no upstream issue, it will make us harder to upgrade in the future.

PR #9824, opened 2023-06-09
chore: process.execPath could contains whitespace

UncleBill:
MF-0000

`"/Users/unclebill/Library/Application Support/fnm/node-versions/v18.7.0/installation/bin/node"`


Jack-Works (approved):
Oh I don't know that!

PR #9841, opened 2023-06-12
refactor: disallow use of React.FC and React.ReactElement

Jack-Works:
mf-0000

React.FC makes it impossible to declare a component with generics. To make our code consistent in style, we should only use the function declaration instead of React.FC.

React.ReactElement is usually wrong. You'd always want to use React.ReactNode. This should only be allowed when cloneElement is used.

### DIFF
--- a/packages/mask/src/plugins/Wallet/SNSAdaptor/GasSettingDialog/GasSetting1559.tsx
+++ b/packages/mask/src/plugins/Wallet/SNSAdaptor/GasSettingDialog/GasSetting1559.tsx
@@ -1,4 +1,4 @@
-import { type FC, memo, useCallback, useEffect, useMemo, useState } from 'react'
+import { memo, useCallback, useEffect, useMemo, useState } from 'react'
 import { useUpdateEffect } from 'react-use'
 import { z as zod } from 'zod'
 import { BigNumber } from 'bignumber.js'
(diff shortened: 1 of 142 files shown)


packages/mask/src/plugins/Wallet/SNSAdaptor/GasSettingDialog/GasSetting1559.tsx
```diff
@@ -199,7 +199,7 @@ export const GasSetting1559: FC<GasSettingProps> = memo(
         const handleConfirm = useCallback(
             (data: zod.infer<typeof schema>) => {
                 onConfirm?.({
-                    gasLimit: data.gasLimit,
+                    gasLimit: data.gasLimit as any,
```
Jack-Works, line 202: I don't know why this isn't a type error before. But data.gasLimit is a string, but gasLimit in onConfirm is number. I won't touch it in this PR. cc @guanbinrui

packages/mask/src/plugins/Wallet/SNSAdaptor/GasSettingDialog/Prior1559GasSetting.tsx
```diff
@@ -167,7 +167,7 @@ export const Prior1559GasSetting: FC<GasSettingProps> = memo(
         const handleConfirm = useCallback(
             (data: zod.infer<typeof schema>) => {
                 onConfirm({
-                    gasLimit: data.gasLimit,
```
Jack-Works, line 170: I don't know why this isn't a type error before. But data.gasLimit is a string, but gasLimit in onConfirm is number. I won't touch it in this PR. cc @guanbinrui

PR #9849, opened 2023-06-15
refactor: create bound modal object

UncleBill:
MF-0000

### DIFF
--- a/packages/shared-base/src/SingletonModal/index.ts
+++ b/packages/shared-base/src/SingletonModal/index.ts
@@ -82,4 +82,13 @@ export class SingletonModal<
             this.onAbort = (error) => reject(error)
         })
     }
+
+    bind() {
+        this.register = this.register.bind(this)
+        this.open = this.open.bind(this)
+        this.close = this.close.bind(this)
+        this.abort = this.abort.bind(this)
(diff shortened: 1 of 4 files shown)


vite.config.ts
```diff
@@ -6,7 +6,7 @@ function createURL(pathToFile: string) {
 
 export default defineConfig({
     test: {
-        include: ['./packages/**/tests/**/*.ts'],
+        include: ['./packages/**/tests/**/*.ts', './packages/**/*.test.ts'],
```
UncleBill, line 9: Putting source code and testing code together, it would be clearer to manage.

guanbinrui (approved):

packages/shared-base/src/SingletonModal/index.ts
```diff
@@ -25,6 +25,14 @@ export class SingletonModal<
     private dispatchClose: ReturnType<T>['close'] | undefined
     private dispatchAbort: ReturnType<T>['abort'] | undefined
 
+    constructor() {
+        this.register = this.register.bind(this)
```
Jack-Works, line 29: this will be cleaner if you use class fields directly. `register = (creator: T | null) => ...`

PR #9893, opened 2023-06-24
chore: quote svg url to preview svg with Raycast

UncleBill:
MF-0000

Raycast can not render a svg if it has unquoted reference url.

## Before
image

## After

image


UncleBill:
Also reported to Raycast team, and asked them if they could fixed this from their side.

cspell.json
```diff
@@ -6,7 +6,7 @@
         "./packages/app/postcss.config.cjs",
         "./packages/app/src/styles/index.css",
         "./packages/app/tailwind.config.cjs",
-        "./packages/dashboard/src/assets/Welcome.splinecode",
```
Jack-Works: rename a file that only differs in cast will break Windows users

Jack-Works (dismissed):

cspell.json
```diff
@@ -6,7 +6,7 @@
...
+        "./packages/dashboard/src/assets/welcome.splinecode",
```
UncleBill:
```suggestion
        "./packages/dashboard/src/assets/Welcome.splinecode",
```

guanbinrui (approved):

PR #9913, opened 2023-06-25
feat: enable lockdown on firefox

Jack-Works:
mf-0000


Jack-Works:
theoretically, this only affects firefox.

PR #9984, opened 2023-07-04
fix: artblocks loading icon

lelenei:
## Description

Closes mf-4402

## Type of change

- [ ] Documentation

## Checklist

### DIFF
--- a/packages/plugins/ArtBlocks/src/SNSAdaptor/Collectible.tsx
+++ b/packages/plugins/ArtBlocks/src/SNSAdaptor/Collectible.tsx
@@ -53,7 +53,7 @@ export function Collectible({ projectId, chainId: projectChainId }: CollectibleP
     const { value, loading, error } = useFetchProject(projectId, chainId)
     const project = value?.projects[0]
 
-    if (!loading)
+    if (loading)
         return (
             <div className={classes.body}>
                 <LoadingBase />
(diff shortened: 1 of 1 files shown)


packages/plugins/ArtBlocks/src/SNSAdaptor/Collectible.tsx
```diff
@@ -46,7 +53,12 @@ export function Collectible({ projectId, chainId: projectChainId }: CollectibleP
     const { value, loading, error } = useFetchProject(projectId, chainId)
     const project = value?.projects[0]
 
-    if (loading) return <LoadingBase />
+    if (!loading)
+        return (
+            <div className={classes.body}>
+                <LoadingBase />
+            </div>
+        )
```
Jack-Works: what about 
```suggestion
    if (loading) return <LoadingBase className={classes.body} />
```

UncleBill: `LoadingBase` is not a semantic name. It's actually an animating loading icon. Applying such block style to it might not suitable.

Recommend using `<LoadingStatus className={classes.body} />` instead. (You can omit loading text by setting `omitText=true`)

Jack-Works (approved):

guanbinrui (approved):

PR #10052, opened 2023-07-12
chore: align wallet icon

UncleBill:
MF-0000

## Before

image

## After
image


packages/icons/general/Wallet.svg
UncleBill, line 1: `Icons.Wallet` is never used previously.

guanbinrui (approved):

PR #10091, opened 2023-07-16
fix: wallet token detail page

UncleBill:
https://mask.atlassian.net/browse/MF-4515
https://mask.atlassian.net/browse/MF-4539
https://mask.atlassian.net/browse/MF-4548
https://mask.atlassian.net/browse/MF-4477
https://mask.atlassian.net/browse/MF-4555


.node-version
```diff
@@ -1 +1 @@
-v18.7.0
```
UncleBill, line 1: Current webpack setup fails on `node<20`

UncleBill:
@guanbinrui PTAL

guanbinrui (approved):

PR #10114, opened 2023-07-18
fix: mf-4561 click on shadowdom cause ui jumping

UncleBill:
https://mask.atlassian.net/browse/MF-4561

### DIFF
--- a/packages/flags/src/flags/index.ts
+++ b/packages/flags/src/flags/index.ts
@@ -13,6 +13,7 @@ export const flags = {
 
     shadowRootInit: {
         mode: '__REACT_DEVTOOLS_GLOBAL_HOOK__' in globalThis || betaOrInsiderOnly ? 'open' : 'closed',
+        delegatesFocus: true,
     } as const satisfies ShadowRootInit,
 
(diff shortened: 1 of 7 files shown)


packages/flags/src/flags/index.ts
```diff
@@ -13,7 +13,6 @@ export const flags = {
 
     shadowRootInit: {
         mode: '__REACT_DEVTOOLS_GLOBAL_HOOK__' in globalThis || betaOrInsiderOnly ? 'open' : 'closed',
-        delegatesFocus: true,
```
UncleBill: Critical change.

Would this break any of your intentions? @Jack-Works

Jack-Works: No

Jack-Works: hmm but you should change it case-by-case

Jack-Works: can you remove this only in the component that has the problem with this flag? cc @UncleBill 

removing this from all things might cause keyboard focus to be unusable again

Jack-Works (approved):

PR #10145, opened 2023-07-23
refactor: ensure plugin does not export from their register

Jack-Works:
mf-0000

why?

Before this PR, plugin is registered with it's main entry point (`@masknet/plugin-x`) but there is a _very bad_ habbit within our developers that adds `export * from './UI'` in the `index.ts` file. This includes React and unnecessary code into the background.

This PR moves entry point to `@masknet/plugin-x/register`. Although not encouraged, but if you need to export from a plugin, you can create a new `index.ts` to do that.


packages/plugins/Approval/src/SNSAdaptor/ApprovalTokenContent.tsx
```diff
@@ -203,7 +203,7 @@ function ApprovalTokenItem(props: ApprovalTokenItemProps) {
         chainId,
     )
 
-    const { value: token } = useFungibleToken(NetworkPluginID.PLUGIN_EVM, spender.tokenInfo.address, undefined, {
+    const { data: token } = useFungibleToken(NetworkPluginID.PLUGIN_EVM, spender.tokenInfo.address, undefined, {
```
Jack-Works, line 206: I found some plugins are not listed in `packages/plugins/tsconfig.json`. After adding that, here comes type error. @UncleBill is this fix correct?

UncleBill: Yes.

PR #10683, opened 2023-08-31
fix: mf-5196 update currency icons

UncleBill:
https://mask.atlassian.net/browse/MF-5195


guanbinrui (approved):

UncleBill:
Oops, it should be https://mask.atlassian.net/browse/MF-5195

PR #10742, opened 2023-09-11
refactor: introduce react persist query client

UncleBill:
MF-0000


packages/shared/src/UI/contexts/PageUIProvider.tsx
```diff
@@ -25,11 +26,25 @@ interface MaskUIRootProps extends React.PropsWithChildren<{}> {
...
+        },
+    },
+}
+
 function MaskUIRoot({ children, useTheme, fallback }: MaskUIRootProps) {
     return (
         <DialogStackingProvider hasGlobalBackdrop={false}>
             <MaskThemeProvider useMaskIconPalette={(theme) => theme.palette.mode} useTheme={useTheme}>
-                <QueryClientProvider client={queryClient}>
+                <PersistQueryClientProvider client={queryClient} persistOptions={persistOptions}>
```
Jack-Works, line 47: I don't think we can add it here. It will store all data in, for example, twitter.com. Other plugins/twitter can access/modify them as well. Please only add it in `PageUIProvider`, not in `SiteUIProvider`.

Jack-Works (changes requested):

packages/shared/src/UI/contexts/PageUIProvider.tsx
UncleBill: This **is** `PageUIProvider`.

Jack-Works: No, this is `MaskUIRoot` which is shared by both providers.

image

packages/shared/src/UI/contexts/PageUIProvider.tsx
```diff
@@ -25,11 +26,25 @@ interface MaskUIRootProps extends React.PropsWithChildren<{}> {
     fallback?: React.ReactNode
 }
 
+const persister = createSyncStoragePersister({
+    storage: window.localStorage,
+})
+// We don't persist all queries but only those have the first key starts with '@@'
```
Jack-Works, line 32: does this means query key like `@@normal-data` will be stored and `normal-data` will not?

packages/shared/src/UI/contexts/PageUIProvider.tsx
UncleBill: Not really, they have their own `<MaskUIRoot>`

packages/shared/src/UI/contexts/PageUIProvider.tsx
UncleBill: > does this means query key like `@@normal-data` will be stored and `normal-data` will not?

Yes, just a magic mark.

packages/shared/src/UI/contexts/PageUIProvider.tsx
Jack-Works: oh sorry I didn't notice that

Jack-Works (approved):

guanbinrui (approved):

PR #10836, opened 2023-09-22
feat: add u8[] support for injected script

Jack-Works:
mf-0000


UncleBill (approved):

UncleBill:
reference https://mask.atlassian.net/browse/MF-5370

PR #11023, opened 2023-10-23
chore: add build-chrome task

UncleBill:
for building production mode locally and quickly

MF-0000

### DIFF
--- a/packages/scripts/src/extension/ci.ts
+++ b/packages/scripts/src/extension/ci.ts
@@ -18,7 +18,14 @@ export const ciBuild: TaskFunction = series(
     zipTo('MaskNetwork.chromium-beta.zip', ManifestFile.ChromiumBetaMV2),
     zipTo('MaskNetwork.chromium-mv3.zip', ManifestFile.ChromiumMV3),
 )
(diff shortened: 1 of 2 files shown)


packages/scripts/src/extension/ci.ts
```diff
@@ -18,7 +18,14 @@ export const ciBuild: TaskFunction = series(
     zipTo('MaskNetwork.chromium-beta.zip', ManifestFile.ChromiumBetaMV2),
     zipTo('MaskNetwork.chromium-mv3.zip', ManifestFile.ChromiumMV3),
 )
+export const buildChrome: TaskFunction = series(
```
Jack-Works, line 21: note, you don't need to add this. `npx gulp build-ci` will run webpack only once, it just makes the zip with a different manifest file many times. All different browser support is done in the runtime.

UncleBill: What I want is building production mode faster locally, `build-chrome` is significantly faster than `build-ci`, zipping takes time!

Jack-Works (approved):

PR #11057, opened 2023-10-31
fix: mf-5559 tip button in reply photo tweet

UncleBill:
https://mask.atlassian.net/browse/MF-5552
https://mask.atlassian.net/browse/MF-5559


UncleBill:
@guanbinrui PTAL

guanbinrui (approved):

PR #11161, opened 2023-11-22
chore: mf-5626 upgrade react-query

UncleBill:
https://mask.atlassian.net/browse/MF-5626


Jack-Works:
sorry I did not notice this! I also tried to upgrade this but not open a PR yet. I have some problems so I just merged #11163 and #11162, after rebasing those PRs, you'll need to handle less type errors.

UncleBill:
> sorry I did not notice this! I also tried to upgrade this but not open a PR yet. I have some problems so I just merged #11163 and #11162, after rebasing those PRs, you'll need to handle less type errors.

Migrating by linting is the right way.

Jack-Works (commented):
you should change `isLoading` to `isPending` first, then change `isInitialLoading` to `isLoading`. that will be easier

packages/web3-hooks/base/src/useHiddenAddressConfig.ts
```diff
@@ -11,6 +11,7 @@ type StorageValueV3 = { hiddenAddresses?: Record<string, string[]> }
 type CurrentStorageValue = StorageValueV3
 type StorageValue = StorageValueV1 | StorageValueV2 | StorageValueV3
 
+type T = UseQueryResult
```
Jack-Works, line 14: do you need to add this one? if a file already imports `UseQueryResult` and uses it somewhere, it's highly likely that it doesn't need this mark.

UncleBill:
We would downgrade react-query to v4 in Firefly instead, since some typing error could be resolved for now.

Jack-Works:
> We would downgrade react-query to v4 in Firefly instead, since some typing error could be resolved for now.

why? I can solve them. I'll do that next week

PR #11171, opened 2023-11-24
Revert "chore: initialize build info at build time"

Jack-Works:
Reverts DimensionDev/Maskbook#11169 mf-0000

cc @UncleBill Hi I want to discuss this change here.

It is a deliberate change to make the build stable. All environmental things should be out of any JS file and be accessed via JSON.

Can you state the reason that you must let them be inlined?


UncleBill:
Intent to make it run on Firefly, like some mocking here:

https://github.com/DimensionDev/Maskbook/compare/firefly-runtime#diff-21489d4ade71339600e0a63b89cccac795ff1ea8bf2c0e3a5b490eb05e51553bR32

These initialization still can be overridden through `setupBuildInfoManually` and `setupBuildInfo`

Jack-Works:
> Intent to make it run on Firefly, like some mocking here:
> 
> [`firefly-runtime`#diff-21489d4ade71339600e0a63b89cccac795ff1ea8bf2c0e3a5b490eb05e51553bR32 (compare)](https://github.com/DimensionDev/Maskbook/compare/firefly-runtime#diff-21489d4ade71339600e0a63b89cccac795ff1ea8bf2c0e3a5b490eb05e51553bR32)
> 
> These initialization still can be overridden through `setupBuildInfoManually` and `setupBuildInfo`

What about adding a `/build-info.json` in firefly? That can solve the problem as well. Or why don't firefly call `setupBuildInfoManually`?

Jack-Works:
hi, @UncleBill please run `setupBuildInfoManually` in Firefly. I'll merge this PR

UncleBill:
We should make plugins side-effect free as well. `registryPlugin()` from plugin make the code hard to trace.

Jack-Works:
> We should make plugins side-effect free as well. `registryPlugin()` from plugin make the code hard to trace.

Yes, it's definitely a thing I want to do. There are still some problems with the TypeScript setup with this approach, so maybe it will be solved later, not now.

PR #11178, opened 2023-11-26
fix: the URL.canParse is still experiment feature

UncleBill:
MF-0000

image

image


Jack-Works (changes requested):
The app package should include the polyfill `packages/polyfills/dist/ecmascript.js`. please don't remove those new API usages.

guanbinrui (approved):

guanbinrui (approved):

PR #11180, opened 2023-11-26
fix: mf-5634 twitter queryIds are outdated

UncleBill:
https://mask.atlassian.net/browse/MF-5634


UncleBill:
BTW, calculation of tweet weight was inaccurate.

guanbinrui (approved):

PR #11193, opened 2023-11-29
refactor: let initial popup requests go through react query

Jack-Works:
mf-0000


packages/shared-base-ui/src/hooks/useValueRef.ts
```diff
@@ -12,6 +13,24 @@ export function useValueRef<T>(ref: ValueRef<T>): Readonly<T> {
...
+        queryFn: async () => {
+            await ref.readyPromise
+            return ref.value
+        },
+        placeholderData: () => ref.value as any,
+        networkMode: 'always',
+    })
+    useEffect(() => {
+        refetch()
+        ref.addListener(() => refetch())
```
UncleBill, line 28: Did you forget to return the clean-up callback?

Jack-Works: nice catch thanks!

PR #11196, opened 2023-12-01
fix: mock for firefly

UncleBill:
MF-0000

### DIFF
--- a/packages/app/src/helpers/createPostInfoContext.ts
+++ b/packages/app/src/helpers/createPostInfoContext.ts
@@ -16,10 +16,10 @@ const emptyValueRef = new ValueRefWithReady<any>()
 
 export function createPostInfoContext(): PostInfo {
     return {
-        site: EnhanceableSite.App,
         nickname: NULL,
         avatarURL: NULL,
         author: NULL,
+        site: EnhanceableSite.App,
         postID: NULL,
 
         get rootNode() {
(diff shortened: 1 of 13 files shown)


packages/flags/src/flags/buildInfo.ts
```diff
@@ -28,7 +28,9 @@ export async function getBuildInfo(): Promise<BuildInfoFile> {
         }
     }
 }
-export let env: BuildInfoFile
+export let env: BuildInfoFile = {
+    channel: 'beta',
```
Jack-Works:
```suggestion
    channel: 'stable',
```
?

and access to this variable before call setupBuildInfo should be a fatal error.

packages/mask/package.json
```diff
@@ -29,7 +29,10 @@
     }
   },
   "exports": {
-    ".": null
+      "./crypto": {
+          "mask-src": "./background/services/crypto/index.ts",
+          "default": "./background/services/crypto/index.js"
+      }
```
Jack-Works: this package should not have any exports. can you explain why you need this?

UncleBill: We would decode encrypted post payload directly in firefly.

Jack-Works: you should use `@masknet/encryption` package directly, which is published on npm. you can see how `packages/app` uses that package as a reference.

packages/flags/src/flags/buildInfo.ts
```diff
@@ -28,7 +28,9 @@ export async function getBuildInfo(): Promise<BuildInfoFile> {
         }
     }
 }
-export let env: BuildInfoFile
+export let env: BuildInfoFile = {
```
Jack-Works: access to this variable before calling setupBuildInfo should be a fatal error, so I'd rather leave it `undefined`.

PR #11208, opened 2023-12-05
feat: add more readonly methods to mask sdk

Jack-Works:
mf-0000

Added more methods as read-only methods so they can be derive-allowed in mask sdk.

### DIFF
--- a/packages/web3-shared/evm/src/helpers/isReadonlyMethodType.ts
+++ b/packages/web3-shared/evm/src/helpers/isReadonlyMethodType.ts
@@ -15,6 +15,18 @@ export const readonlyMethodType = [
     EthereumMethodType.ETH_ESTIMATE_GAS,
     EthereumMethodType.ETH_CALL,
     EthereumMethodType.ETH_GET_LOGS,
+    EthereumMethodType.ETH_FEE_HISTORY,
+    EthereumMethodType.ETH_GET_BLOCK_RECEIPTS,
+    EthereumMethodType.ETH_GET_BLOCK_TRANSACTION_COUNT_BY_HASH,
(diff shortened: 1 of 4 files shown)


packages/web3-shared/evm/src/helpers/isReadonlyMethodType.ts
```diff
@@ -15,6 +15,18 @@ export const readonlyMethodType = [
...
+    EthereumMethodType.ETH_GET_BLOCK_TRANSACTION_COUNT_BY_HASH,
+    EthereumMethodType.ETH_GET_BLOCK_TRANSACTION_COUNT_BY_NUMBER,
+    EthereumMethodType.ETH_GET_PROOF,
+    EthereumMethodType.ETH_GET_STORAGE_AT,
+    EthereumMethodType.ETH_GET_TRANSACTION_BY_BLOCK_HASH_AND_INDEX,
+    EthereumMethodType.ETH_GET_TRANSACTION_BY_BLOCK_NUMBER_AND_INDEX,
+    EthereumMethodType.ETH_GET_TRANSACTION_BY_HASH,
+    EthereumMethodType.ETH_GET_UNCLE_COUNT_BY_BLOCK_HASH,
+    EthereumMethodType.ETH_GET_UNCLE_COUNT_BY_BLOCK_NUMBER,
+    EthereumMethodType.ETH_SYNCING,
```
Jack-Works, line 29: @guanbinrui you may want to check this list

Jack-Works: please give an explicit approve if you think this is ok @guanbinrui

guanbinrui (approved):

PR #11212, opened 2023-12-08
fix: mf-5746 allow to disable setting for calendar

UncleBill:
https://mask.atlassian.net/browse/MF-5746

Would disable calendar setting on Firefly


Jack-Works (approved):

UncleBill:
@guanbinrui PTAL

PR #11227, opened 2023-12-26
chore: patch @tanstack/query-async-storage-persister to fix performance problem

Jack-Works:
mf-0000

workaround for https://github.com/TanStack/query/issues/6489

cc @UncleBill


UncleBill (approved):

UncleBill:
@guanbinrui We should pick this patch to `2.24`

PR #11229, opened 2023-12-26
fix: decode data value might not be BigNumber

UncleBill:
https://mask.atlassian.net/browse/MF-5746
https://mask.atlassian.net/browse/MF-5866


guanbinrui (approved):

UncleBill:
@guanbinrui Please pick to 2.24.

Also fixes https://mask.atlassian.net/browse/MF-5903

PR #11273, opened 2024-01-11
fix: mf-5270 check response in interaction page

UncleBill:


UncleBill:
@guanbinrui PTAL

guanbinrui (approved):

PR #11296, opened 2024-01-17
fix: mf-5994 check if ValueRef is ready

UncleBill:
https://mask.atlassian.net/browse/MF-5994

### DIFF
--- a/packages/shared-base-ui/src/hooks/useValueRef.ts
+++ b/packages/shared-base-ui/src/hooks/useValueRef.ts
@@ -8,7 +8,7 @@ function getServerSnapshot(): never {
 export function useValueRef<T>(ref: ValueRef<T>): T {
     if ('readyPromise' in ref) {
         const ref2 = ref as ValueRefWithReady<T>
-        if (!ref2.nowReady) use(ref2.readyPromise)
+        if (!ref2.ready) use(ref2.readyPromise)
     }
     return useSyncExternalStore(
         (f) => ref.addListener(f),


packages/plugin-infra/src/manager/site-adaptor.ts
```diff
@@ -34,7 +34,16 @@ useActivatedPluginsSiteAdaptor.visibility = {
 }
 
 export function useIsMinimalMode(pluginID: string) {
-    return useValueRef(minimalModeSub).includes(pluginID)
```
Jack-Works: why do you need this change? useValueRef already handles `readyPromise`.

UncleBill: The result was `false` by default, that's arbitrary.

UncleBill: @Jack-Works Any other questions?

Jack-Works: the `use(ref.readyPromise)` should suspend the render, you should not get a "default" value before the `readyPromise` is resolved.

UncleBill:
```ts
        if (!ref2.nowReady) use(ref2.readyPromise)
```
Should be 
```ts
        if (!ref2.ready) use(ref2.readyPromise)
```

Jack-Works: nice catch

Jack-Works (approved):

PR #11325, opened 2024-01-28
fix: trusted types

Jack-Works:
mf-0000


packages/mask/dashboard/Dashboard.tsx
```diff
@@ -70,9 +70,10 @@ export default function Dashboard() {
         <>
             <CssBaseline />
             {GlobalCss}
-            {process.env.NODE_ENV === 'development' ?
+            {/* https://github.com/TanStack/query/issues/5417 */}
+            {/* {process.env.NODE_ENV === 'development' ?
                 <ReactQueryDevtools buttonPosition="bottom-right" />
-            :   null}
+            :   null} */}
```
UncleBill, line 76: Why comment out `ReactQueryDevtools`? @Jack-Works

Jack-Works: it is still crashing see https://github.com/TanStack/query/issues/5417

PR #11400, opened 2024-02-15
ci: try macos-14

UncleBill:
MF-0000

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -20,7 +20,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: macos-14
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/compile.yml
```diff
@@ -20,7 +20,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
```
Jack-Works, line 23: what happened 👀

UncleBill: https://twitter.com/onevcat/status/1757787865793413470 It's said that `macos-14` is significantly faster.

But our CI runs out of the memory. Maybe we should allocate more to the node job.

Jack-Works: try a bigger `--max_old_space_size`?

PR #11440, opened 2024-02-26
chore: upgrade node to 21

Jack-Works:
mf-0000


swkatmask:
This change fails the `dev` command. (Already using node 21)

Jack-Works:
> This change fails the `dev` command. (Already using node 21)

Fixing, sorry!

Jack-Works:
already fixed!

PR #11470, opened 2024-02-29
feat: support gasLimit and gasPrice edit in send tx

Jack-Works:
mf-0000 cc @guanbinrui


Jack-Works:
before

a

after 
b

packages/mask/popups/pages/Wallet/Interaction/index.tsx
```diff
@@ -273,6 +273,13 @@ const Interaction = memo((props: InteractionProps) => {
 
     const [{ loading: confirmLoading }, handleConfirm] = useAsyncFn(
         async (paymentToken: string, approveAmount: string, gasConfig: GasConfig | undefined) => {
+            console.log(
+                'Sending',
+                currentRequest.request.arguments.method,
+                currentRequest.request.arguments.params,
+                'mixing with gasConfig:',
+                gasConfig,
+            )
```
guanbinrui:
```suggestion
```

packages/mask/popups/pages/Wallet/Interaction/index.tsx
```diff
@@ -306,28 +306,35 @@ const Interaction = memo((props: InteractionProps) => {
...
                     )
                 }
 
+                if (currentRequest.request.arguments.method === EthereumMethodType.ETH_SEND_TRANSACTION) {
+                    if (params[0].type === '0x0') {
+                        delete params[0].type
+                        delete params[0].gasPrice
+                    }
+                }
+                console.log('Finalized parameters', params)
```
guanbinrui: Remove logs

packages/mask/popups/modals/GasSettingModal/GasSettingDialog.tsx
```diff
@@ -132,6 +138,11 @@ export const GasSettingDialog = memo<GasSettingDialogProps>(function GasSettingM
         return
     }, [gasPrice, gasOptions, isSupport1559])
 
+    const gasLimitError = useMemo(() => {
+        if (isGreaterThanOrEqualTo(gasLimit, MinGasLimit) && isLessThanOrEqualTo(gasLimit, MaxGasLimit)) return
```
nuanyang233: Why set a max limit🤔? Is there any reference for this value?

packages/mask/popups/pages/Wallet/Interaction/index.tsx
```diff
@@ -306,28 +306,35 @@ const Interaction = memo((props: InteractionProps) => {
...
                                 ...(gasConfig ?
                                     mapValues(omit(gasConfig, 'gasOptionType'), (value, key) => {
                                         if (key === 'gasCurrency' || !value) return
                                         return web3_utils.toHex(value)
                                     })
                                 :   {}),
-                                gas: web3_utils.toHex(new BigNumber(gasConfig?.gas ?? x.gas).toString()),
-                                chainId: web3_utils.toHex(x.chainId),
-                                nonce: web3_utils.toHex(x.nonce),
+                                gasLimit: web3_utils.toHex(new BigNumber(gasConfig?.gas ?? param.gas).toString()),
```
nuanyang233: Changing field names may cause rpc calls to fail.

packages/mask/popups/modals/GasSettingModal/GasSettingDialog.tsx
Jack-Works: https://etherscan.io/blocks

A block can only use 30,000,000 gas. tx exceed this limit won't be packed into the block

packages/mask/popups/pages/Wallet/Interaction/index.tsx
Jack-Works: the old name `gas` does not work at all. only `gasLimit` works.

packages/mask/popups/modals/GasSettingModal/GasSettingDialog.tsx
nuanyang233: OK, but it's not necessarily the same case with other chains. eg: https://bscscan.com/blocks.

Jack-Works: > Each block has a target size of 15 million gas

https://ethereum.org/developers/docs/gas

so 15,000,000 gas is a reasonable max value, which only allows 1 tx in 1 (minimal) block.

Jack-Works: > OK, but it's not necessarily the same case with other chains. eg: [bscscan.com/blocks](https://bscscan.com/blocks).

hmmm ok, I'll remove that limit

guanbinrui (approved):

nuanyang233 (approved):

PR #11478, opened 2024-03-04
fix: mf-6071 don't persistent cache my persona info

swkatmask:
https://github.com/DimensionDev/Maskbook/pull/11193

https://mask.atlassian.net/browse/mf-6071

### DIFF
--- a/packages/shared/src/hooks/usePersonaContext.ts
+++ b/packages/shared/src/hooks/usePersonaContext.ts
@@ -25,7 +25,7 @@ function usePersonaInformation(
     queryOwnedPersonaInformation?: (initializedOnly: boolean) => Promise<PersonaInformation[]>,
 ) {
     const { data: personas = EMPTY_LIST, refetch } = useQuery({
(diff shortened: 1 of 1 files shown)


guanbinrui (approved):

packages/shared/src/hooks/usePersonaContext.ts
```diff
@@ -25,7 +25,7 @@ function usePersonaInformation(
     queryOwnedPersonaInformation?: (initializedOnly: boolean) => Promise<PersonaInformation[]>,
 ) {
     const { data: personas = EMPTY_LIST, refetch } = useQuery({
-        queryKey: ['@@queryOwnedPersonaInformation(false)'],
+        queryKey: ['my-own-persona-info'],
```
Jack-Works, line 28: we cache this because we need to fix the loading speed. your fix is wrong, you should change all other references to this key to `@@queryOwnedPersonaInformation(false)`

PR #11482, opened 2024-03-05
fixup! fix: mf-6071 don't persistent cache my persona info (#11478)

swkatmask:
https://mask.atlassian.net/browse/mf-6071


Jack-Works (changes requested):
packages/mask/popups/pages/Personas/AccountDetail/index.tsx (line 86 and 157)

Jack-Works (approved):

PR #11499, opened 2024-03-15
fix: correct some configuration for Metis

swkatmask:
MF-0000


swkatmask:
Should pick to hotfix branch as well. @guanbinrui

guanbinrui (approved):

PR #11506, opened 2024-03-20
fix: connected personas might be outdate

swkatmask:
MF-0000 FW-0000
https://mask.atlassian.net/browse/FW-512


swkatmask:
https://github.com/DimensionDev/Maskbook/pull/11508

PR #11539, opened 2024-03-31
refactor: set background-image for icons on Firefox

swkatmask:
MF-0000 FW-0000

### DIFF
--- a/packages/icons/utils/internal.js
+++ b/packages/icons/utils/internal.js
@@ -60,6 +60,15 @@ export function __createIcon(name, variants, intrinsicSize = [24, 24]) {
             // To align icon center.
             fontSize: 0,
         }
+        // background image can't be correctly resolved on Firefly in ShadowDOM.
+        // This is a workaround.
+        // https://mask.atlassian.net/browse/FW-564
+        if (iconStyle.backgroundImage && navigator.userAgent.includes('Firefox')) {
+            iconProps.style = {
+                ...iconProps.style,
+                backgroundImage: iconStyle.backgroundImage,
+            }
+        }
         if (hasClickHandler) {
             iconProps['aria-hidden'] = false
             iconProps['role'] = 'button'


packages/polyfills/types/env.d.ts
```diff
@@ -1,6 +1,7 @@
 declare module NodeJS {
     interface ProcessEnv {
         readonly NODE_ENV: 'development' | 'production' | 'test'
+        readonly RUNTIME_HOST: 'Firefly' | string
```
guanbinrui: Not used?

packages/icons/utils/internal.js
```diff
@@ -60,6 +60,15 @@ export function __createIcon(name, variants, intrinsicSize = [24, 24]) {
             // To align icon center.
             fontSize: 0,
         }
+        // background image can't be correctly resolved on Firefly in ShadowDOM.
+        // This is a workaround.
+        // https://mask.atlassian.net/browse/FW-564
+        if (iconStyle.backgroundImage && navigator.userAgent.includes('Firefox')) {
```
guanbinrui, line 66: CC @Jack-Works

Jack-Works: why this can fix the problem? looks like it just repeat the backgroundImage property?

swkatmask: Not the same.

With `sx` prop, `mui` creates stylesheet and injects the style in it. In this way, Firefox can not correctly resolve the image. While inline style can fix that.

It might be a bug of Firefox that the Stylesheet in the ShadowDOM doesn't inherit the `baseURI` of the document.

image

Jack-Works: can you write this in the comment? it will be easier to tell future people why you'r doing this

swkatmask: > can you write this in the comment? it will be easier to tell future people why you'r doing this

It's already written above.

Jack-Works: I mean in the code comment

PR #11561, opened 2024-04-15
chore: remove react-query patch

Jack-Works:
mf-0000

see #11227 for why we have this patch

seems like they fixed the problem, it's running fast now!

@UncleBill please also have a try


guanbinrui (approved):

swkatmask:
LGTM

PR #11613, opened 2024-05-13
fix: fw-555 wallet address is required to claim redpacket

swkatmask:
https://mask.atlassian.net/browse/fw-555


swkatmask:
https://github.com/DimensionDev/Maskbook/pull/11614

PR #11649, opened 2024-05-24
refactor: get rid of forwardRef

Jack-Works:
mf-0000

https://react.dev/blog/2024/04/25/react-19#ref-as-a-prop I'll update Firefly to ready for this change later


swkatmask:
might break Firefly 🧐

Jack-Works:
> might break Firefly 🧐

I'll upgrade it to https://nextjs.org/blog/next-15-rc

PR #11656, opened 2024-05-29
fix: add sentry dsn

swkatmask:
MF-0000 FW-0000


.github/workflows/compile.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       WEB3_CONSTANTS_RPC: ${{ secrets.WEB3_CONSTANTS_RPC }}
-      MASK_SENTRY_DSN: ${{ secrets.MASK_SENTRY_DSN }}
+      MASK_SENTRY_DSN: https://9da9e7dcc47cd0b229067760202b13e5@sentry.firefly.land/9
```
Jack-Works, line 26: why not set it in the CI secret?

PR #11657, opened 2024-05-29
fix: mf-6014 don't replace tweets that have embed images

Jack-Works:
close mf-6014
close mf-6242

this was made by @swkatmask in #11425 but for an unknown reason unmerged.
I have tested that this fixes the issue


swkatmask (approved):

Jack-Works:
cc @guanbinrui you may want to pick this

PR #11670, opened 2024-06-13
perf: prevent non-gas-consuming calls invoke gas translators

Jack-Works:
mf-0000


packages/web3-shared/evm/src/helpers/isGasConsumingMethodType.ts
```diff
@@ -0,0 +1,8 @@
+import { EthereumMethodType } from '../types/index.js'
+
+export const gasConsumingMethodType = [
```
Jack-Works, line 3: cc @guanbinrui is this list ok?

PR #11693, opened 2024-07-16
refactor: slim maskbook for firefly

swkatmask:


Jack-Works:
looks like you removed almost everything. looks like not a correct PR. close for now

swkatmask:
> looks like you removed almost everything. looks like not a correct PR. close for now

You don't even ask me.

This is done on purpose. We would like to fork and maintain a slim version of Maskbook with critical code only, and this fork would never be merged into Maskbook main branch.

swkatmask:
This is done on purpose.

Jack-Works:
> > looks like you removed almost everything. looks like not a correct PR. close for now
> 
> You don't even ask me.
> 
> This is done on purpose. We would like to fork and maintain a slim version of Maskbook with critical code only, and this fork would never be merged into Maskbook main branch.

sorry, this PR looks like a mistake so I closed it. What if the main branch made some changes, is there any plan to follow the upstream, or it will keep forked forever? If it's the latter case, I'd suggest just copy code to ff

swkatmask:
It's OK.

> or it will keep forked forever?

Yes, it will.

> If it's the latter case, I'd suggest just copy code to ff

We might do it in the future, then Firefly can get rid of Maskbook in that day.

/cc @guanbinrui

PR #11743, opened 2024-08-15
refactor: remove unused files

Jack-Works:
mf-0000

continue of #11742 cc @nuanyang233


nuanyang233:
Do you still need these codes? @swkatmask

Jack-Works:
> Do you still need these codes? @swkatmask

They're unreachable from the user's perspective (and always reachable in the git history).

PR #11753, opened 2024-08-20
feat: mf-6093 migrate rss3 api

swkatmask:
https://mask.atlassian.net/browse/MF-6094
https://mask.atlassian.net/browse/MF-6333


guanbinrui (approved):

packages/plugins/RSS3/src/SiteAdaptor/FeedFilters.tsx
```diff
@@ -0,0 +1,94 @@
...
+    },
+    networks: {
+        display: 'flex',
+        gap: theme.spacing(1, 1.5),
+        flexWrap: 'wrap',
+    },
+}))
+
+export const FeedFilters = memo(
+    forwardRef<HTMLDivElement, BoxProps>(function FeedFilters(props: BoxProps, ref) {
```
Jack-Works, line 46: forwardRef has been deprecated. see https://react.dev/blog/2024/04/25/react-19#ref-as-a-prop

swkatmask: But mui is not compatible with that yet.

swkatmask: image

Jack-Works: it should be like this:

```js
// old
export default forwardRef<HTMLDivElement, BoxProps>(function (props: BoxProps, ref) {})

// new
export default function({ ...props, ref }: BoxProps & { ref: ... }) {}
```

if this does not work, that will be strange

PR #11767, opened 2024-08-26
ci: add tag to disable ci cache

swkatmask:
MF-0000 FW-0000


.github/workflows/compile.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         with:
             registry: true
       - name: Restore Webpack cache
```
Jack-Works, line 48: I have discussed this with @UncleBill and I think you can remove the whole cache section.

PR #11771, opened 2024-08-28
fix(Twitter): update query id for UserByScreenName

swkatmask:
MF-0000 FW-0000

### DIFF
--- a/packages/web3-providers/src/Twitter/apis/getUserByScreenName.ts
+++ b/packages/web3-providers/src/Twitter/apis/getUserByScreenName.ts
@@ -22,7 +22,9 @@ const features = {
 }
 
 async function createRequest(screenName: string) {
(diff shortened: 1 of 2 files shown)


guanbinrui (approved):

packages/web3-providers/src/Twitter/apis/getUserByScreenName.ts
```diff
@@ -22,7 +22,9 @@ const features = {
 }
 
 async function createRequest(screenName: string) {
-    const url = urlcat(twitterDomainMigrate('https://x.com/i/api/graphql/sLVLhk0bGj3MVFEKTdax1w/UserByScreenName'), {
+    // cspell:disable-next-line
+    // Yka-W8dz7RaEuQNkroPkYw is queryId for UserByScreenName
+    const url = urlcat('https://x.com/i/api/graphql/Yka-W8dz7RaEuQNkroPkYw/UserByScreenName', {
```
Jack-Works, line 27: you should wrap this with `twitterDomainMigrate`

PR #11772, opened 2024-08-28
refactor: remove some more unused files

swkatmask:


packages/shared-base/src/i18n/instance.ts
```diff
@@ -6,10 +6,10 @@ import { debounce, type DebouncedFunc } from 'lodash-es'
 export const i18NextInstance = i18n.default || i18n
 if (process.env.NODE_ENV === 'development') {
     if (Reflect.get(globalThis, '__mask_shared_base__')) {
-        throw new Error('@masknet/shared-base initialized twice. Please check your code.')
```
Jack-Works, line 9: this check has its meaning. multiple initialize this package causes mysterious bugs.

swkatmask: Do worry, just remove it for Firefly. It keep throwing in Firefly, too many errors get logged in console.

guanbinrui (approved):

PR #11846, opened 2024-10-10
chore: upgrade lint

Jack-Works:
mf-0000


packages/mask/shared-ui/locale/en-US.po
```diff
@@ -274,7 +274,7 @@ msgstr ""
 
 #: dashboard/pages/CreateMaskWallet/CreateWalletForm/index.tsx:136
 #: popups/modals/ChangePaymentPasswordModal/index.tsx:88
-#: popups/pages/Wallet/SetPaymentPassword/index.tsx:248
+#: popups/pages/Wallet/SetPaymentPassword/index.tsx:245
```
swkatmask, line 277: Let's disable code position from extraction, to reduce changes in `po` file.

Jack-Works: > Let's disable code position from extraction, to reduce changes in `po` file.

is that possible? I don't find how on lingui document

swkatmask: It is.

refer to https://github.com/DimensionDev/firefly.mask.social/blob/8917e2a740000ae19d76fedebc0bb8b9df74e7c4/lingui.config.js#L34-L35

PR #11875, opened 2024-10-21
feat: add a temp solution for adding trader storage in popup

Jack-Works:
mf-0000


Jack-Works:
this already picked in to develop

PR #11898, opened 2024-11-05
fix: mf-6464 setState in ref cause infinite update

swkatmask:
https://mask.atlassian.net/browse/mf-6464


Jack-Works:
hi, I fix this without reverting the change in #11899 so I'm closing this. thank you!

PR #11961, opened 2024-11-29
fix: x image with correct size and no dup

Jack-Works:
mf-0000


swkatmask (approved):

swkatmask:
@guanbinrui  Please pick this patch to v2.29.0 as well.

PR #12113, opened 2025-02-14
fix: mf-6623 support snapshot v2

swkatmask:
https://mask.atlassian.net/browse/mf-6623
https://mask.atlassian.net/browse/mf-6624
https://mask.atlassian.net/browse/MF-6627

### DIFF
--- a/packages/plugins/Snapshot/src/SiteAdaptor/ResultCard.tsx
+++ b/packages/plugins/Snapshot/src/SiteAdaptor/ResultCard.tsx
@@ -102,6 +102,7 @@ function Content() {
     const votes = useVotes(identifier)
     const { results } = useResults(identifier, proposal)
     const { classes, cx } = useStyles()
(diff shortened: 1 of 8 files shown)


packages/plugins/Snapshot/src/SiteAdaptor/ResultCard.tsx
```diff
@@ -102,6 +102,7 @@ function Content() {
     const votes = useVotes(identifier)
     const { results } = useResults(identifier, proposal)
     const { classes, cx } = useStyles()
+    console.log({ results })
```
Jack-Works: don't forget to remove this

swkatmask: Nice catch, thnaks!

guanbinrui (approved):

PR #12183, opened 2025-04-07
fix: mf-6672 request identity permission

swkatmask:
https://mask.atlassian.net/browse/mf-6672

### DIFF
--- a/packages/mask/shared/helpers/checkAndRequestIdentityPermission.ts
+++ b/packages/mask/shared/helpers/checkAndRequestIdentityPermission.ts
@@ -1,0 +1,10 @@
+import type { Manifest } from 'webextension-polyfill'
+
+export async function checkAndRequestPermission() {
+    const contained = await browser.permissions.contains({ permissions: ['identity'] })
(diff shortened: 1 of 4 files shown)


guanbinrui (approved):

packages/mask/shared/helpers/checkAndRequestIdentityPermission.ts
```diff
@@ -0,0 +1,10 @@
+import type { Manifest } from 'webextension-polyfill'
+
+export async function checkAndRequestPermission() {
+    const contained = await browser.permissions.contains({ permissions: ['identity'] })
```
Jack-Works, line 4: this helper should check for the environment like `requestPermissionFromExtensionPage` did:

```ts
    if (!isEnvironment(Environment.ExtensionProtocol) || isEnvironment(Environment.ManifestBackground)) {
        // The User Activation limitation is from Firefox
        throw new Error(
            'browser.permissions.request can only be called after a User Activation and from a chrome-extension:// protocol.',
        )
    }
```

PR #12158, opened 2025-03-07
Integrate Weavevm Bundler

codingknite:
MF-0000

PR #12236, opened 2025-05-23
[Release] Hotfix 2.33.2 => 2.33.3 (patch)

guanbinrui:
MF-0000


Jack-Works:
@guanbinrui looks like this PR is containing #12158, should we merge that first?

guanbinrui:
> @guanbinrui looks like this PR is containing #12158, should we merge that first?

I think we can merge #12158 later. We're currently testing on this PR.

PR #12250, opened 2025-07-01
feat: allow firefly share x oauth with mask

Jack-Works:
mf-0000


Jack-Works:
I'll merge this since it is used in #12281

PR #12272, opened 2025-09-02
feat: add support for farcaster social network integration

swkatmask:
MF-0000 FW-0000


packages/web3-providers/src/Session/Session.ts
```diff
@@ -0,0 +1,27 @@
...
+    constructor(
+        public type: SessionType,
+        public profileId: string,
+        public token: string,
+        public createdAt: number,
+        public expiresAt: number,
+    ) {}
+
+    serialize(): `${SessionType}:${string}` {
+        const body = JSON.stringify({
```
Jack-Works, line 13: In case this is part of some protocol, or will be signed, you may need `json-stable-stringify`

packages/shared-base/src/errors.ts
```diff
@@ -0,0 +1,41 @@
+export class AbortError extends Error {
```
Jack-Works, line 1: looks like this class is never constructed. why not make it simple, `isAbortError`?

packages/mask/popups/pages/Personas/ConnectFirefly/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+                if (AbortError.is(error)) return
+
+                // if login timed out, let the user refresh the QR code
+                if (error instanceof TimeoutError || error instanceof FireflyBindTimeoutError) {
+                    showSnackbar(<Trans>This QR code is longer valid. Please scan a new one to continue.</Trans>)
+                    return
+                }
+
+                // failed to patch the signer
+                if (error instanceof FarcasterPatchSignerError) throw error
```
Jack-Works, line 63: this class is never constructed, in which case this will be hit?

Jack-Works (changes requested):

Jack-Works:
The failed test is broken from nodejs 24.5.0 (known good) to 24.7.0 (known bad). I will investigate this.

Jack-Works:
Hi you should pick #12276 to fix the test failure

PR #12281, opened 2025-09-26
feat: mf-6735 integrate privy wallet support

swkatmask:
- add Privy provider and API integration for wallet management
- support Privy wallets in MaskWalletProvider and wallet hooks
- enable login and wallet creation via X (Twitter) account with auth timeout handling
- adapt transaction requests to handle Privy wallet signing
- update cspell dictionaries and ignore reg exps for privy related domains
- adjust UI components for importing/creating wallets with Privy option

https://mask.atlassian.net/browse/mf-6735


packages/mask/.webpack/flags.ts
```diff
@@ -37,6 +37,8 @@ export interface BuildFlags {
     sourceMapPreference?: boolean | string
     /** @default true */
     sourceMapHideFrameworks?: boolean | undefined
+    FIREFLY_X_CLIENT_ID?: string
+    FIREFLY_X_CLIENT_SECRET?: string
```
guanbinrui, line 41: Remove those env variables from flags.

swkatmask: Removed

packages/mask/background/services/helper/firefly.ts
```diff
@@ -0,0 +1,33 @@
+import { PersistentStorages } from '@masknet/shared-base'
+
+export async function loginFireflyViaTwitter() {
+    const data = await browser.storage.local.get('firefly_x_oauth')
+    const oauth: Record<string, string> = data.firefly_x_oauth
```
Jack-Works, line 6: What if the data is empty? Please handle this case.

packages/mask/content-script/utils/shadow-root/ContentScriptGlobalProvider.tsx
```diff
@@ -1,37 +1,51 @@
...
 import { PersistQueryClientProvider } from '@tanstack/react-query-persist-client'
+import { Suspense } from 'react'
+import { createPortal } from 'react-dom'
 import { queryPersistOptions } from '../../../shared-ui/utils/persistOptions.js'
-import { LinguiProviderHMR, SharedContextProvider } from '@masknet/shared'
-import { useSiteThemeMode } from '@masknet/plugin-infra/content-script'
 import { useMaskSiteAdaptorMixedTheme } from '../../components/useMaskSiteAdaptorMixedTheme.js'
 
 export function ContentScriptGlobalProvider(children: React.ReactNode) {
-    return jsxCompose(
```
Jack-Works, line 17: Why remove the usage of `jsxCompose`?

packages/mask/popups/Popup.tsx
```diff
@@ -108,23 +108,35 @@ export default function Popups() {
         throttle: 10000,
     })
 
-    return jsxCompose(
```
Jack-Works, line 111: Why remove the usage of `jsxCompose`? This helper function is good for maintenance.

packages/mask/content-script/utils/shadow-root/ContentScriptGlobalProvider.tsx
swkatmask: It is suitable to compose single-child jsx, if there are multiple children, it might end up xml style. You can see there is a `<PrivySetup/>` alongside `<SharedContextProvider />` as `RootWeb3ContextProvider's` children

Besides, it causes every component **renders twice**, once for the original, once for the cloned one.

packages/mask/content-script/utils/shadow-root/ContentScriptGlobalProvider.tsx
```diff
@@ -1,37 +1,51 @@
...
-        <MaskThemeProvider useMaskIconPalette={useSiteThemeMode} useTheme={useMaskSiteAdaptorMixedTheme} />,
-    )(
-        cloneElement,
+    const jsx =
         process.env.NODE_ENV === 'development' ?
             <>
                 {/* https://github.com/TanStack/query/issues/5417 */}
                 {createPortal(<ReactQueryDevtools buttonPosition="bottom-right" />, document.body)}
                 {children}
             </>
```
Jack-Works, line 34: Multiple children should go here. Does that answer your question?

packages/mask/content-script/utils/shadow-root/ContentScriptGlobalProvider.tsx
Jack-Works: What do you mean by renders twice? Does that mean the JSX tree is being duplicated in the final rendering tree?

packages/mask/content-script/utils/shadow-root/ContentScriptGlobalProvider.tsx
swkatmask: I think xml style would be more convenient, it's inconvenient to add siblings. Like if you prepend a sibling before `<DialogStackingProvider />`, should we move all descendant nodes to the callback parameter, and keep descendant nodes in xml or compose them with another `jsxCompose()`?

I think `xml` is better to manage and present hierarchical relationship.

To avoid deep indent, we can reduce the indent width, that's why I ignored prettier for this, or we can group those providers by dimensions, like `<IOProvider/>` groups io-related providers, etc.

Anyway, I can revert to `jsxCompose` if you insist.

packages/mask/content-script/utils/shadow-root/ShadowRootAttachPointRoot.tsx
```diff
@@ -1,24 +1,21 @@
...
-            anchorOrigin={{ vertical: 'top', horizontal: 'right' }}
-            children={null!}
-            offsetY={Sniffings.is_facebook_page ? 80 : undefined}
-        />,
-    )(
-        cloneElement,
-        <>
-            <CSSVariableInjector />
-            {children}
-        </>,
```
swkatmask, line 22: Let lines of code, and clear hierarchical relation at a glance.

packages/mask/content-script/utils/shadow-root/ContentScriptGlobalProvider.tsx
swkatmask: > What do you mean by renders twice? Does that mean the JSX tree is being duplicated in the final rendering tree?

Sorry, it might be that I misread it at the time.

packages/mask/.webpack/flags.ts
```diff
@@ -37,8 +37,6 @@ export interface BuildFlags {
     sourceMapPreference?: boolean | string
     /** @default true */
     sourceMapHideFrameworks?: boolean | undefined
-    FIREFLY_X_CLIENT_ID?: string
```
Jack-Works, line 40: Why are you removing those variables?

packages/mask/popups/pages/Wallet/Interaction/WalletSignRequest.tsx
```diff
@@ -21,13 +21,25 @@ type ParsedSigningMessage = string | EIP4361Message | undefined
...
     const request = currentRequest.request.arguments
 
     const { Message } = useWeb3State(NetworkPluginID.PLUGIN_EVM)
     useEffect(() => setConfirmVerb(<Trans>Sign</Trans>), [])
 
     setConfirmAction(async () => {
+        if (privyWallet?.isConnected) {
+            const provider = await privyWallet.getEthereumProvider()
+            const result = await provider.request(request)
+            await Message!.updateMessage(currentRequest.ID, {
```
guanbinrui:
```suggestion
            await Message?.updateMessage(currentRequest.ID, {
```

packages/mask/popups/pages/Wallet/Interaction/TransactionRequest.tsx
```diff
@@ -79,14 +80,19 @@ const approveParametersType = [
...
+    const wallet = useWallet()
+    const privyWallet = usePrivyWallet(wallet?.address)
     const { currentRequest: request, setConfirmAction, paymentToken, setPaymentToken } = props
     const { classes, cx } = useStyles()
     const [gasConfig, _setGasConfig] = useState<GasConfig | undefined>()
     const [approvedAmount, setApproveAmount] = useState('')
     const { Message, TransactionFormatter } = useWeb3State(NetworkPluginID.PLUGIN_EVM)
     const { chainId } = useChainContext<NetworkPluginID.PLUGIN_EVM>()
 
+    assert(Message, 'Wallet transaction request: Message not found')
```
guanbinrui:
```suggestion
```

PR #12325, opened 2025-12-15
fix: mf-6765 update json-stable-stringify and use default export

swkatmask:
https://mask.atlassian.net/browse/mf-6765

fixup https://github.com/DimensionDev/Maskbook/pull/12259


guanbinrui (approved):

packages/shared-base/src/helpers/ValueRef.ts
```diff
@@ -1,7 +1,7 @@
 // All imports must be deferred. This file loads in the very early stage.
 import defer * as lodash from 'lodash-es'
 import defer * as constants from '../constants.js'
-import defer * as stringify from 'json-stable-stringify'
+import stringify from 'json-stable-stringify'
```
Jack-Works, line 4: Hi, the comment writes that all imports must be deferred!

swkatmask: I did notice that.
But we need a solution for https://mask.atlassian.net/browse/MF-6765.
All code are loaded from disk, I think cost of loading this package is small enough.

PR #12340, opened 2026-01-20
feat(mask): fw-6767 add firefly desktop sync for twitter cookies

swkatmask:
https://mask.atlassian.net/browse/fw-6767


packages/web3-hooks/base/src/useWallet.ts
```diff
@@ -8,11 +8,12 @@ import { useWallets } from './useWallets.js'
  * @param pluginID
  * @returns
  */
-export function useWallet() {
+export function useWallet(address?: string) {
```
Jack-Works, line 11: in this case, you can write `useWallets().find(x => isSameAddress(x.address, address))`

swkatmask: That's why I edited this.

guanbinrui (approved):

PR #12410, opened 2026-07-03
fix: script.src in bundle

Jack-Works:
mf-0000


Jack-Works:
@copilot create a new PR, pick this patch for the develop branch

guanbinrui (approved):

PR #12460, opened 2026-09-07
fix(tips): stop treating GoPlus contract_address flag as malicious

swkatmask:
## Summary

- Tips flagged vitalik.eth (any EIP-7702 delegated EOA) as a "malicious address". Root cause: GoPlus marks any address with code as `contract_address: "1"` — since EIP-7702, delegated EOAs have code — and `useRecipientValidate` scanned **every** field of the security response for `"1"`, conflating "is a contract" with "is malicious".
- Extract `isMaliciousAddress()` into the provider layer (`GoPlusLabs`): it ignores `contract_address` (contract-ness is judged locally via `getAddressType`, which already handles 7702 correctly) and `data_source`, and checks the remaining risk flags.
- `Tips/useRecipientValidate` and `GoPlusLabs.checkIfAddressIsScam` (used by ScamWarning for Solana/Tron) now share it.

## Known limitation (pre-existing, unchanged)

`number_of_malicious_contracts_created` is a count field: values ≥ 2 don't equal the string `'1'` and go undetected; value `'1'` triggers and happens to be semantically correct.

## Verification

- Unit tests for `isMaliciousAddress` (web3-providers vitest): all-`"0"` → false; `contract_address:"1"` alone → false (the 7702 case); `phishing_activities:"1"` → true. 3/3 pass.
- Live: rebuilt dev extension, opened Tips on a @VitalikButerin post (mainnet, recipient vitalik.eth). TanStack query devtools confirm the `["go-plus","address-security",1,0xd8dA…]` query resolved (stale, 1 observer, 20-item data) and the malicious-address warning no longer renders.
- `tsc --noEmit` clean on `web3-providers` and `plugins/Tips`; eslint clean on all changed files.

### DIFF
--- a/packages/web3-providers/src/GoPlusLabs/index.ts
+++ b/packages/web3-providers/src/GoPlusLabs/index.ts
@@ -10,6 +10,9 @@ import { SecurityMessages } from './rules.js'
 import { getAllMaskDappContractInfo } from '../helpers/getAllMaskDappContractInfo.js'
 import { fetchJSON } from '../helpers/fetchJSON.js'
 import type { AuthorizationAPI, PhishingSiteResponse, SecurityAPI } from '../entry-types.js'
+import { isMaliciousAddress } from './isMaliciousAddress.js'
+
+export { isMaliciousAddress } from './isMaliciousAddress.js'
 
 function checkInWhitelist(chainId: ChainId, address: string) {
     const { WHITE_LISTS } = getGoPlusLabsConstants(chainId)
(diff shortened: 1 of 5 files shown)


packages/web3-providers/src/GoPlusLabs/index.ts
```diff
@@ -10,6 +10,9 @@ import { SecurityMessages } from './rules.js'
 import { getAllMaskDappContractInfo } from '../helpers/getAllMaskDappContractInfo.js'
 import { fetchJSON } from '../helpers/fetchJSON.js'
 import type { AuthorizationAPI, PhishingSiteResponse, SecurityAPI } from '../entry-types.js'
+import { isMaliciousAddress } from './isMaliciousAddress.js'
+
+export { isMaliciousAddress } from './isMaliciousAddress.js'
```
guanbinrui, line 15: No reexport.

guanbinrui (approved):

